PR #42, opened 2023-07-07
Document release process

dmathieu:
This is extremely simple, but documenting it aims to avoid having doubt about things that would be necessary but not documented.


dmathieu:
Once this is approved/merged, I will cut a new release for 0.2.0.

marclop (approved):

PR #44, opened 2023-07-11
tracing: Add `documents` transaction tag

marclop:
Tags the `docappender.flush` transaction with the number of documents that are sent in the bulk request. The tag name is `documents`.


dmathieu (approved):
Ah, I was looking for the way to do that last week. Thanks.

PR #51, opened 2023-08-01
feat: support otel metrics for existing metrics

### LINKED ISSUE #37: Support OTel metrics
## Description

Currently, the go-docappender metrics are recorded as unexported `int64` metrics. It would be ideal if we started using an OTel meter to record those metrics and have them exported by the OTel exporter that's configured in the clients.

https://github.com/elastic/go-docappender/blob/b714d6181871607fe9ccb48bfb59ddc0b57318f5/appender.go#L63-L74

We could do this as a breaking change and remove the `Stats()` method, or attempt to do this as a non-breaking change, both have advantages and disadvantages.

### Breaking change

If we do it as a breaking change, we will need to find a way to keep the `apm-server` libbeat metric reporting contract in place.

https://github.com/elastic/apm-server/blob/78ec846c77bffe42553c9e116b1f647a575fee60/internal/beater/beater.go#L716-L770

### Non Breaking change

Another option would be to keep the `Stats()` method and maintain both metric recording mechanisms to avoid creating a breaking change, complexity can be _abstracted_ away, but can't be eliminated.

kyungeunni:
closes #37 

Support otel metrics for the existing metrics so it can be exported by the client

### DIFF
--- a/appender_test.go
+++ b/appender_test.go
@@ -79,9 +79,9 @@ func TestAppender(t *testing.T) {
 		},
 	))
 
-	indexerAttrs := []attribute.KeyValue{
+	indexerAttrs := attribute.NewSet(
 		attribute.String("a", "b"), attribute.String("c", "d"),
-	}
+	)
 
 	indexer, err := docappender.New(client, docappender.Config{
 		FlushInterval:    time.Minute,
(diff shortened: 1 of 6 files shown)


appender_test.go
```diff
@@ -117,6 +133,45 @@ loop:
 		AvailableBulkRequests: 10,
 		BytesTotal:            bytesTotal,
 	}, stats)
+
+	var rm metricdata.ResourceMetrics
+	assert.NoError(t, rdr.Collect(context.Background(), &rm))
+	var asserted int
+	assertCounter := func(metric metricdata.Metrics, count int64, attrs attribute.Set) {
+		asserted++
+		// assert.Equal(t, "s", metric.Unit)
```
axw, line 142:
```suggestion
```

appender_test.go
```diff
@@ -117,6 +133,45 @@ loop:
...
+			assertCounter(metric, stats.FailedServer, wantAttrs)
+		case docappender.METRIC_NAME_DOCS_INDEXED:
+			assertCounter(metric, stats.Indexed, wantAttrs)
+		case docappender.METRIC_NAME_TOO_MANY_REQ:
+			assertCounter(metric, stats.TooManyRequests, wantAttrs)
+		case docappender.METRIC_NAME_AVAILABLE_BULK_REQ:
+			assertCounter(metric, stats.AvailableBulkRequests, wantAttrs)
+		case docappender.METRIC_NAME_BYTES_TOTAL:
+			assertCounter(metric, stats.BytesTotal, wantAttrs)
+		}
```
axw, line 172: Should we add a `default` case here, to make sure we don't miss checking any metrics that we add in the future?

appender_test.go
```diff
@@ -117,6 +133,45 @@ loop:
...
+		asserted++
+		// assert.Equal(t, "s", metric.Unit)
+		counter := metric.Data.(metricdata.Sum[int64])
+		for _, dp := range counter.DataPoints {
+			assert.Equal(t, count, dp.Value)
+			assert.Equal(t, attrs, dp.Attributes)
+		}
+	}
+	wantAttrs := attribute.NewSet(indexerAttrs...)
+	for _, metric := range rm.ScopeMetrics[0].Metrics {
```
axw, line 150: What this doesn't tell us is that all of the metrics are reported - only _if_ they're recorded, that their values are correct.

Should we use the https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest?

axw (commented):
Looking good! Just minor issues.

appender_test.go
kyungeunni: I think the purpose of this test is to assert that the values are recorded correctly, but not necessarily reported by the exporter, what do you think? 

Also, I had a look at the `[metricdatatest](https://pkg.go.dev/go.opentelemetry.io/otel/sdk/metric/metricdata/metricdatatest) package, and it seems it is a bit more complicated to use that what we already have here.

axw: I missed the "asserted" check. It looks good now with the `default` case.
Thanks for looking into `metricdatatest`, sounds fine.

appender_test.go
```diff
@@ -117,6 +133,54 @@ loop:
...
+			// expect this metric name but no assertions done
+			// as it's histogram and it's checked elsewhere
+		case "elasticsearch.flushed.latency":
+			// expect this metric name but no assertions done
+			// as it's histogram and it's checked elsewhere
+		default:
+			unexpectedMetrics = append(unexpectedMetrics, metric.Name)
+		}
+	}
+	assert.Equal(t, []string{}, unexpectedMetrics)
```
axw, line 182:
```suggestion
	assert.Empty(t, unexpectedMetrics)
```

axw (approved):
Thanks for the updates, LGTM!

PR #53, opened 2023-08-07
feat: dimensional metrics for `elasticsearch.events.processed` metric

kyungeunni:
Instead of creating several metrics having status in names, create a single metric with different attributes/dimentions.

Before -> after
- `elasticsearch.events.processed` -> `elasticsearch.events.processed {status: "Success"}`
- `elasticsearch.failed.client.count ` ->  `elasticsearch.events.processed {status: "FailedClient"}`
- `elasticsearch.failed.server.count` -> `elasticsearch.events.processed {status: "FailedServer"}`
- `elasticsearch.failed.too_many_reqs` -> `elasticsearch.events.processed {status: "TooMany"}
`

PR #54, opened 2023-08-08
feat: dimensional metrics for `elasticsearch.events.processed` metric 

kyungeunni:
Instead of creating several metrics having status in names, create a single metric with different attributes/dimensions.

Before -> after
- `elasticsearch.events.processed` -> `elasticsearch.events.processed {status: "Success"}`
- `elasticsearch.failed.client.count ` ->  `elasticsearch.events.processed {status: "FailedClient"}`
- `elasticsearch.failed.server.count` -> `elasticsearch.events.processed {status: "FailedServer"}`
- `elasticsearch.failed.too_many_reqs` -> `elasticsearch.events.processed {status: "TooMany"}
`

The outcome of it is the same with the #53, but it uses synchronous instruments instead.

### DIFF
--- a/appender.go
+++ b/appender.go
@@ -38,19 +38,12 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
-type ProcessedStatus string
-
 var (
 	// ErrClosed is returned from methods of closed Indexers.
 	ErrClosed = errors.New("model indexer closed")
 
 	errMissingIndex = errors.New("missing index name")
 	errMissingBody  = errors.New("missing document body")
-
-	success      ProcessedStatus = "Success"
-	failedClient ProcessedStatus = "FailedClient"
-	failedServer ProcessedStatus = "FailedServer"
-	tooMany      ProcessedStatus = "TooMany"
 )
 
 // Appender provides an append-only API for bulk indexing documents into Elasticsearch.
@@ -89,7 +82,6 @@ type Appender struct {
 	errgroupContext       context.Context
 	cancelErrgroupContext context.CancelFunc
 	telemetryAttrs        attribute.Set
-	processedEventAttrSet map[ProcessedStatus]attribute.Set
 	metrics               metrics
 	mu                    sync.Mutex
 	closed                chan struct{}
@@ -133,13 +125,6 @@ func New(client *elasticsearch.Client, cfg Config) (*Appender, error) {
 		}
 	}
 
-	baseAttrs := cfg.MetricAttributes.ToSlice()
-	processedEventAttrSet := map[ProcessedStatus]attribute.Set{
-		success:      attribute.NewSet(append(baseAttrs, attribute.String("status", "Success"))...),
-		failedClient: attribute.NewSet(append(baseAttrs, attribute.String("status", "FailedClient"))...),
-		failedServer: attribute.NewSet(append(baseAttrs, attribute.String("status", "FailedServer"))...),
-		tooMany:      attribute.NewSet(append(baseAttrs, attribute.String("status", "TooMany"))...),
-	}
 	ms, err := newMetrics(cfg)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


appender.go
```diff
@@ -38,12 +38,19 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+type ProcessedStatus string
```
axw, line 41:
```suggestion
type processedStatus string
```

I don't think this needs to be exported - let's avoid exporting things unless needed.

appender.go
```diff
@@ -82,6 +89,7 @@ type Appender struct {
 	errgroupContext       context.Context
 	cancelErrgroupContext context.CancelFunc
 	telemetryAttrs        attribute.Set
+	processedEventAttrSet map[ProcessedStatus]attribute.Set
```
axw, line 92: In all cases where we update the metric, the status is known statically. So if we're going to create reusable attribute sets, I think we should store them as separate fields and avoid the map lookup overhead. Something like this:

```go
type Appender struct {
    telemetryAttrs                     attribute.Set
    statusSuccessAttributeSet attribute.Set
    statusFailedClientAttributeSet attribute.Set
    ...
}
```

However, thinking about it a bit more, it might not be worth the complexity as the metric is only updated on flush. Let's keep it simple for now and just pass in a new `WithAttributes` option for each update.

appender_test.go
```diff
@@ -411,7 +411,7 @@ func TestAppenderFlushMetric(t *testing.T) {
 			if !ignoreCount {
 				assert.Equal(t, count, int(dp.Count))
 			} else {
-				assert.Greater(t, int(dp.Count), count)
+				assert.GreaterOrEqual(t, int(dp.Count), count)
```
kyungeunni, line 414: I debugged it locally with the main branch, and it also quite often emits 2 counts here, it should be ok to expect greater or equal than 2 if I'm not mistaken.

appender.go
```diff
@@ -240,6 +256,14 @@ func (a *Appender) Add(ctx context.Context, index string, document io.Reader) er
 	return nil
 }
 
+func (a *Appender) addProcessedCount(delta int64, lm *int64, m metric.Int64Counter, status ProcessedStatus) {
+	// legacy metric
+	atomic.AddInt64(lm, delta)
+
+	attr := a.processedEventAttrSet[status]
+	m.Add(context.Background(), delta, metric.WithAttributeSet(attr))
+}
```
axw, line 265: As above, let's keep it simple and not create a reusable attribute set for the `status` attribute. Instead I think we should remove `addProcessedCount`, and add a `... metric.AddOption` parameter to `addCount`. Then we can call it like this:

`a.addCount(int64(n), &a.tooManyRequests, a.metrics.docsIndexed, metric.WithAttributes(attribute.String("status", "TooMany")))`

WDYT?

appender.go
axw: I think we can just delete the type though - see other comments.

appender.go
kyungeunni: Thanks for the suggestion. I went with this approach and decided not to do that as we discussed that we wanted to avoid the overhead of creating an attribute set every time. But I like this approach more if the overhead is not a big concern here. I've addressed it in the later commit.

metric.go
```diff
@@ -92,39 +88,19 @@ func newMetrics(cfg Config) (metrics, error) {
...
-		},
-		{
-			name:        "elasticsearch.failed.server.count",
-			description: "The number of docs failed to get indexed with server error(status_code >= 500).",
-			p:           &ms.docsFailedServer,
-		},
 		{
 			name:        "elasticsearch.events.processed",
-			description: "The number of docs indexed successfully.",
+			description: "Number of APM Events flushed to Elasticsearch. Dimensions are used to report the project ID, success or failures",
```
axw, line 101:
```suggestion
			description: "Number of APM Events flushed to Elasticsearch. Attributes are used to report separate counts for different outcomes - success, client failure, etc.",
```

metric.go
```diff
@@ -56,7 +52,7 @@ type counterMetric struct {
 	p           *metric.Int64Counter
 }
 
-func newMetrics(cfg Config) (metrics, error) {
+func newMetrics(cfg Config) (*metrics, error) {
```
axw, line 55: Is this necessary? I think we should either revert this, or update the `Appender.metrics` field type to `*metrics`

axw (approved):
LGTM, just a couple of nitpicks. Thanks!

metric.go
kyungeunni: Oh good catch, it isn't necessary. It's leftover from the previous experiment. thanks!

Issue #57, opened 2023-08-28
Reduce network usage by using filter paths in bulk requests


kruskall:
This is inspired by the work of the Ingest team during the last offsite.

Because we are using very little of the total response sent by ES, we could use filter path to reduce the payload and potentially improve performance

PR #62, opened 2023-08-29
perf: use filter_path to ignore unused bulk response fields

kruskall:
Closes https://github.com/elastic/go-docappender/issues/57


carsonip (approved):

axw:
@kruskall did you happen to do any benchmarking? Curious if it has a noticeable impact, given that we have optimised the parsing.

PR #81, opened 2023-11-21
feat: add FlushTimeout config and use context timeout for flush requests

kruskall:
The ES client is using the err group context when flushing which means the bulk index request could hang forever.
This PR adds a FlushTimeout option which is passed to the flush context, ensuring a deadline is always present.
For backward compatibility and to avoid changing behaviour the default value is no timeout.

### DIFF
--- a/appender_test.go
+++ b/appender_test.go
@@ -348,6 +348,31 @@ func TestAppenderFlushInterval(t *testing.T) {
 	}
 }
 
+func TestAppenderFlushTimeout(t *testing.T) {
+	done := make(chan struct{}, 1)
+	defer close(done)
+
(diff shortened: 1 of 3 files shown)


appender_test.go
```diff
@@ -348,6 +348,31 @@ func TestAppenderFlushInterval(t *testing.T) {
...
+	})
+	require.NoError(t, err)
+	defer indexer.Close(context.Background())
+
+	addMinimalDoc(t, indexer, "logs-foo-testing")
+
+	assert.Eventually(t, func() bool {
+		return indexer.Stats().BulkRequests == 1
+	}, 10*time.Second, 10*time.Millisecond)
+}
```
simitt, line 374: Maybe I am overlooking something, but how is this testing the `FlushTimeout`? It is only checking that there was a bulk request.

simitt (approved):

PR #89, opened 2023-12-04
logging: Single log line per unique error type

marclop:
Logs a single consolidated line per unique error, instead of logging a line for each failed document in a bulk request. Drastically reduces the number of log lines when a bulk request with a high number of failed documents happens.

### DIFF
--- a/appender.go
+++ b/appender.go
@@ -313,6 +313,10 @@ func (a *Appender) flush(ctx context.Context, bulkIndexer *bulkIndexer) error {
 	}
 	var docsFailed, docsIndexed, tooManyRequests, clientFailed, serverFailed int64
 	docsIndexed = resp.Indexed
+	var failedCount map[BulkIndexerResponseItem]int
+	if len(resp.FailedDocs) > 0 {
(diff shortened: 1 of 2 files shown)


appender.go
```diff
@@ -326,21 +330,19 @@ func (a *Appender) flush(ctx context.Context, bulkIndexer *bulkIndexer) error {
 			if info.Status >= 500 {
 				serverFailed++
 			}
-			// NOTE(axw) error type and reason are included
-			// in the error message so we can observe different
-			// error types/reasons when logging is rate limited.
```
axw, line 331: I think the NOTE is still relevant - log messages may still be duplicated across multiple requests, and a rate limited logger could still be used to suppress them.

What is the benefit to this change vs. using a rate-limited logger like https://pkg.go.dev/go.uber.org/zap@v1.26.0/zapcore#NewSamplerWithOptions?

marclop: What I like about this approach is that it quantifies the number of documents in the log message, and that's included in the log line.

PR #91, opened 2023-12-05
refactor: compress events in batches to support document retries

kruskall:
This is the groundwork needed to support per-document retries.

The idea is to compress events one by one and store the offsets of the buffer in the indexer. Because the response returns failed events in order to extract the failed event from the compressed payload.

For performance reasons, we're not literally compressing events one by one if they are too small to avoid the following issue: https://github.com/elastic/go-docappender/issues/94
If the request is still too smal on flush timeout we avoid compression completely.

Additionally, `io.Reader` has been changed to `io.WriterTo` for clarity. Note that this is not a breaking change as users of this library are passing an `io.WriterTo`.

### DIFF
--- a/appender.go
+++ b/appender.go
@@ -130,7 +130,7 @@ func New(client *elasticsearch.Client, cfg Config) (*Appender, error) {
 	}
 	available := make(chan *bulkIndexer, cfg.MaxRequests)
 	for i := 0; i < cfg.MaxRequests; i++ {
-		available <- newBulkIndexer(client, cfg.CompressionLevel)
+		available <- newBulkIndexer(client, cfg.CompressionLevel, cfg.FlushBytes/2)
 	}
 	if cfg.Logger == nil {
 		cfg.Logger = zap.NewNop()


endorama (commented):
I reviewed the PR and it looks good to me!

I'm not familiar enough with this code area to have a clear understanding of the performance consequences at scale, are there any benchmarks that I can refer to about the impact of these changes?

bulk_indexer.go
```diff
@@ -162,17 +158,23 @@ type bulkIndexerItem struct {
...
-		return err
+	if _, err := item.Body.WriteTo(&b.buf); err != nil {
+		return fmt.Errorf("failed to write bulk indexer item: %w", err)
+	}
+	if _, err := b.buf.Write([]byte("\n")); err != nil {
+		return fmt.Errorf("failed to write newline: %w", err)
 	}
-	if _, err := b.writer.Write([]byte("\n")); err != nil {
-		return err
+	if b.gzipw != nil && b.buf.Len() > 512*1024 {
```
marclop, line 173: I would prefer if this is configured in the `newBulkIndexer`. This will allow the compress threshold to be a function of `Config.FlushBytes` (could be half or 3/4 of it). We could call it `compressThreshold`.

marclop (commented):
The changes look good. The benchmarks look similar in my machine. I didn't see any measurable differences.

kyungeunni (approved):

PR #99, opened 2024-01-04
feat: add document retries

kruskall:
~Blocked by https://github.com/elastic/go-docappender/pull/100 (to make testing easier)~

~Add document retries logic. Followup to https://github.com/elastic/go-docappender/pull/91 (important to understand how the compression logic changed)~

~### Compression off:~

~In this case we simply iterate over the array and submit the failed events to the buffer. We can count them using newlines.~

~### Compression on:~

~Retrieve the offset of the failed event and decompress the "gzip container". Loop over the array and submit the failed event to buffer.~

### Testing

This PR adds tests to cover as mush edge cases as possible: one event for each gzip container, multiple events per gzip container, all events in a gzip container. Tests also verify the retry logic twice to ensure there are no conflicts between flushes.

Update:

### Changes:

the approach was simplified: iterate over the array and resubmit the failed events to the buffer. For gzip we read in batches to avoid decompress the whole request.

### DIFF
--- a/bulk_indexer.go
+++ b/bulk_indexer.go
@@ -339,8 +339,8 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
 						buf = buf[:n]
 					}
 
-					startIdx := Indexnth(buf, startln-seen, '\n') + 1
-					endIdx := Indexnth(buf, endln-seen, '\n') + 1
+					startIdx := indexnth(buf, startln-seen, '\n') + 1
+					endIdx := indexnth(buf, endln-seen, '\n') + 1
 
 					// If the end newline is not in the buffer read more data
 					for endIdx == 0 {
@@ -351,13 +351,13 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
 						buf = buf[:len(buf)+n]
 
 						// try again to find the end newline
-						endIdx = Indexnth(buf, endln-seen, '\n') + 1
+						endIdx = indexnth(buf, endln-seen, '\n') + 1
 					}
 
 					b.writer.Write(buf[startIdx:endIdx])
 				} else {
-					startIdx := Indexnth(b.copyBuf, startln, '\n') + 1
-					endIdx := Indexnth(b.copyBuf, endln, '\n') + 1
+					startIdx := indexnth(b.copyBuf, startln, '\n') + 1
+					endIdx := indexnth(b.copyBuf, endln, '\n') + 1
 
 					b.writer.Write(b.copyBuf[startIdx:endIdx])
 				}
@@ -378,9 +378,9 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
 	return resp, nil
 }
 
-// Indexnth returns the index of the nth instance of sep in s.
+// indexnth returns the index of the nth instance of sep in s.
 // It returns -1 if sep is not present in s or nth is 0.
-func Indexnth(s []byte, nth int, sep rune) int {
+func indexnth(s []byte, nth int, sep rune) int {
 	if nth == 0 {
 		return -1
 	}


bulk_indexer.go
```diff
@@ -240,9 +255,74 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
+			if b.gzipw != nil {
+				gr, err := gzip.NewReader(bytes.NewReader(copyBuf))
+				if err != nil {
+					return resp, fmt.Errorf("failed to decompress request payload: %w", err)
+				}
+				defer gr.Close()
+
+				seen := 0
+
+				buf := make([]byte, 1024)
```
marclop, line 273: Could we reuse this buffer after its contents are written to the `gzip.Writer`? We're creating at least a 1KB `[]byte` for each failed item. If that buffer is expanded, I imagine it will be more efficient to recycle the byte slice after writing to the gzip compressor.

bulk_indexer.go
```diff
@@ -240,9 +255,74 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
+				n, _ := gr.Read(buf[:1024])
```
marclop, line 274: Wouldn't it suffice to `gr.Read(buf)`?

bulk_indexer.go
```diff
@@ -240,9 +255,74 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
+
+				for end == 0 {
+					// Add more capacity (let append pick how much).
+					buf = append(buf, 0)[:len(buf)]
+
+					n, _ := gr.Read(buf[len(buf):cap(buf)])
+					buf = buf[:len(buf)+n]
+
+					end = Indexnth(buf, endlnIdx-seen, '\n') + 1
+				}
```
marclop, line 293: The current test cases don't seem to cover the contents of the two `for` clauses. Would you mind adding a test or two where the bulk item is large enough to require more capacity?

marclop (commented):
I've noticed that the tests have `MaxRequests: 1`, which is why the tests are passing. However, in a more realistic scenario with `MaxRequests > 1`, the documents stored in the local `bulk_indexer` won't be flushed to Elasticsearch until the bulk indexer is cycled through the channel.

https://github.com/elastic/go-docappender/blob/c8c4e54dbd8820190e465c0811122cb6ffbde24d/appender.go#L470-L479

We should prevent that from happening. I think we may need to introduce a new channel for bulk_indexers that have items from previous 429 failures, and treat it with higher priority than `a.available`. There are some edge cases we need to handle (like closing the appender), but it would be better than simply sending the bulk_indexer with cached items back to the channel like empty indexers.

Another concern is that there's currently no limit on the number of times `429s` will be retried. In theory, the entire buffer could be filled with documents that have not been indexed due to `429s`, which would cause only 1 new event to be sent in each subsequent flush. Have you given any thought on how we could configure an upper limit of consecutive retries?

bulk_indexer.go
```diff
@@ -132,8 +137,10 @@ func init() {
 	})
 }
 
-func newBulkIndexer(client *elasticsearch.Client, compressionLevel int) *bulkIndexer {
+func newBulkIndexer(client *elasticsearch.Client, compressionLevel int, maxRetryDoc int) *bulkIndexer {
 	b := &bulkIndexer{client: client}
+	b.maxDocumentRetry = maxRetryDoc
+	b.retryCounts = make(map[int]int)
```
marclop, line 143: I think it looks a bit more consistent and concise this way. WDYT?
```suggestion
func newBulkIndexer(client *elasticsearch.Client, compressionLevel int, maxDocRetry int) *bulkIndexer {
	b := &bulkIndexer{
		client:           client,
		maxDocumentRetry: maxDocRetry,
		retryCounts:      make(map[int]int),
	}
```

bulk_indexer.go
```diff
@@ -220,8 +227,13 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
 		}
 	}
 
-	copyBuf := make([]byte, b.buf.Len())
-	copy(copyBuf, b.buf.Bytes())
+	if b.maxDocumentRetry != 0 {
```
marclop, line 230: Can it be negative?
```suggestion
	if b.maxDocumentRetry > 0 {
```

bulk_indexer.go
```diff
@@ -257,52 +271,82 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
 		return resp, fmt.Errorf("error decoding bulk response: %w", err)
 	}
 
-	buf := make([]byte, 1024)
+	if b.maxDocumentRetry != 0 {
+		buf := make([]byte, 1024)
+
```
marclop, line 276:
```suggestion
		// Eliminate previous retry counts that aren't present in the last bulk
		// request response.
```

bulk_indexer.go
```diff
@@ -257,52 +271,82 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
+		for k := range b.retryCounts {
+			found := false
+			for _, res := range resp.FailedDocs {
+				if res.Position == k {
+					found = true
```
marclop, line 281: Do we need to keep iterating if the position matches?
```suggestion
					found = true
					break
```

bulk_indexer.go
```diff
@@ -257,52 +271,82 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
-				start := Indexnth(copyBuf, startlnIdx, '\n') + 1
-				end := Indexnth(copyBuf, endlnIdx, '\n') + 1
-
-				b.writer.Write(copyBuf[start:end])
+				tmp = append(tmp, res)
 			}
+		}
 
-			b.itemsAdded++
+		if len(tmp) != 0 {
```
marclop, line 348: Can it be negative?
```suggestion
		if len(tmp) > 0 {
```

config.go
```diff
@@ -55,6 +55,9 @@ type Config struct {
 	// If MaxRequests is less than or equal to zero, the default of 10 will be used.
 	MaxRequests int
 
+	// MaxDocumentRetries holds the maximum number of document retries
+	MaxDocumentRetries int
```
marclop, line 59: Should this default to 3?

bulk_indexer.go
```diff
@@ -257,52 +271,82 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
-					return resp, fmt.Errorf("failed to decompress request payload: %w", err)
+				count := b.retryCounts[res.Position] + 1
+				if count > b.maxDocumentRetry {
+					tmp = append(tmp, res)
+					continue
 				}
-				defer gr.Close()
 
-				seen := 0
+				b.retryCounts[b.itemsAdded] = count
```
marclop, line 301: I think a couple of comments would go a long way:
```go
				// Increment 429 count for the positions found.
				count := b.retryCounts[res.Position] + 1
				if count > b.maxDocumentRetry {
					tmp = append(tmp, res)
					continue
				}
				// Since some items may have succeeded, counter positions need
				// to be updated to match the next current buffer position.
				b.retryCounts[b.itemsAdded] = count
```

I like the retry tracking algorithm, it's neat!

marclop (commented):
> To be honest, I think this is fine. We already changed quite a lot of things because of concerns around complexity. If ES is returning 429s there are a lot of events buffered/coming in. I don't think retried events are gonna stay "idle" for a long time.

I don't think it's acceptable to do all of this retry work and then return a potentially full `bulkIndexer` back to the `available` channel, where it may or may not be flushed in the future.

While it may be the case that 429s tend to happen on higher throughputs and the `bulkIndexer` may not remain in that channel for longer, it won't be flushed on `appender.Close()`, leading to those buffered events to be completely lost.

Could you please add a test that ensures that previously buffered 429s are flushed once more on `indexer.Close()`?

In the case where 429s are a considerable % of the total documents in a bulk request, say 30-40%, we'd leave a bunch of half-full (or almost entirely full if 429 % is even higher) `bulkIndexers` in that `available` channel.

> I think we should get rid of the "active" channel entirely and let the runtime handle the channels but that should go in a separate PR.

Could you elaborate on this? I'm not following your reasoning and how the runtime would be of any help to us in this case.

simitt:
The `bench-diff` check is green, but looking at the [logs](https://github.com/elastic/go-docappender/actions/runs/7783017661/job/21220601839?pr=99), some benchmark errors happened; please take a look and either fix the issue or loop in the automation team, in case it's not legit errors.

appender.go
```diff
@@ -179,18 +176,29 @@ func (a *Appender) Close(ctx context.Context) error {
...
+		return err
 	}
-	return a.errgroup.Wait()
+	close(a.available)
+	for bi := range a.available {
+		if err := a.flush(context.Background(), bi); err != nil {
+			return fmt.Errorf("failed to flush events on close: %w", err)
+		}
+	}
+	return nil
```
marclop, line 201: Returning early will cause the non-empty `bulkIndexers` to not be flushed. What do you think about the snippet below?
```go
	// Wait until all active indexers have been flushed.
	err := a.errgroup.Wait()
	close(a.available)
	for bi := range a.available {
		a.errgroup.Go(func() error {
			if e := a.flush(context.Background(), bi); e != nil {
				return fmt.Errorf("failed to flush events on close: %w", e)
			}
			return nil
		})
	}
	return errors.Join(err, a.errgroup.Wait())
```

bulk_indexer.go
marclop: That was what my comment was about, rather than `!=`, it seems more appropriate to use the `>` operator.

marclop (commented):
Changes mostly LGTM, I think we need to retry flushing even if the error group returns an error.

> > Could you elaborate on this? I'm not following your reasoning and how the runtime would be of any help to us in this case.
> 
> Ideally, we would just read from the channel and let the runtime switch between them, removing the `active` channel completely. The flush interval would then cause a flush request for the bulk indexers that still have events in the buffer. The issue of bulk indexers remaining in a "limbo" would not apply anymore because they would all be active.

I still don't follow what you meant.

Also, have you run some benchmarks using the `tilt` environment and monitoring how the metrics look? I think it's worth doing so before merging.

bulk_indexer.go
simitt: +1 on using `>` for better readability when only dealing with positive numbers.

appender.go
simitt: Could you open a follow up issue then to tackle this? With adding retry behavior we are moving more towards reliability vs. dropping events, so the ask to change the behavior for flushing non empty bulk indexers seems totally fair.

simitt:
>> Also, have you run some benchmarks using the tilt environment and monitoring how the metrics look? I think it's worth doing so before merging.

>I only run benchmarks using the appender benchmarks in the repo but can you clarify this so we are aligned ? Are you asking to test the performance of the retry code or the overhead on the "normal operation" (requests succeed) ?

I do not want to speak for Marc, but IMO we should have some numbers on how this impacts overall performance. I would expect 
- no performance change if there are no retries
- retry behavior to add only a reasonable amount of overhead, but slowing down the processing of new events

Testing this e2e under load in a dev environment is essential before starting the promotion process.

marclop:
@simitt, that's right.

@kruskall, I'll leave it up to you whether you want to test this as part of the dependency update or as part of the PR. If we go the dependency promotion path, we may approve the PR without any regression testing. I understand the retry behavior is opt-in rather than opt-out, but that's what regression testing is for.

marclop (approved):
LGTM. It'd be great to test this when either as part of the dependency update or before we merge.

bulk_indexer.go
```diff
@@ -240,9 +272,128 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
+					tmp = append(tmp, res)
+					continue
+				}
+
+				// Since some items may have succeeded, counter positions need
+				// to be updated to match the next current buffer position.
+				b.retryCounts[b.itemsAdded] = count
+
+				if b.gzipw != nil {
+					gr, err := gzip.NewReader(bytes.NewReader(b.copyBuf))
```
axw, line 323: It looks like we're decompressing and scanning through the request body for every failed item. Should we be decompressing once, and scanning through incrementally? e.g. if you have failures at index 1 and 8, then first skip to the index 1, then skip from index 1 to index 8 without decompressing and reprocessing up to index 1.

axw (commented):
If we don't need to export Indexnth then it would be better not to, to keep the public API tidy. The way the document indexes are found looks a bit inefficient, but we could improve it in a follow up since it should not be on a hot path.

inge4pres (approved):

bulk_indexer.go
axw: I don't mean decompress the whole payload at once, I meant create the decompressor once and stream through the results. AFAICS there's no need to call `gzip.NewReader` more than once.

Instead you could lazily create it when seeing the first 429 result, wrap it with [bufio.NewScanner](https://pkg.go.dev/bufio#NewScanner), and stream through the lines, skipping non-429 result lines as needed.

PR #100, opened 2024-01-04
feat: add appender config for compression threshold

kruskall:
No change in behaviour, just making the config more explicit


carsonip (approved):
lgtm

PR #122, opened 2024-02-08
perf: avoid rereading the buffer from the start for noncompressed requests

kruskall:
Instead of rereading the buffer from the beginning start from where we left off with the previous failed item.


marclop (approved):
LGTM!

PR #123, opened 2024-02-08
perf: reuse gzip reader and avoid rereading buffer for compressed requests

kruskall:
Similar to the other change for uncompressed requests, do not reread
the buffer from the beginning and start where we left off.

The gzip reader is being reused and more data is read lazily.

Depends on https://github.com/elastic/go-docappender/pull/122

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,27 @@
+version: 2
+updates:
+  # Go dependencies
+  - package-ecosystem: "gomod"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    groups:
+      otel:
+        patterns:
+        - "go.opentelemetry.io/*"
+      go-agent:
+        patterns:
+        - "go.elastic.co/apm*"
+  # GitHub actions
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    reviewers:
+      - "elastic/observablt-ci"
+    schedule:
+      interval: "weekly"
+      day: "sunday"
+      time: "22:00"
+    groups:
+      github-actions:
+        patterns:
+          - "*"
--- a/.github/renovate.json
+++ b/.github/renovate.json
@@ -1,9 +1,0 @@
-{
-  "$schema": "https://docs.renovatebot.com/renovate-schema.json",
-  "extends": [
-    "config:base"
-  ],
-  "automerge": true,
-  "automergeStrategy": "squash",
-  "gomodTidy": true
-}


bulk_indexer.go
```diff
@@ -322,35 +336,37 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
 
 					// If the end newline is not in the buffer read more data
-					for endIdx == 0 {
-						// Add more capacity (let append pick how much).
-						buf = append(buf, 0)[:len(buf)]
+					if endIdx == 0 {
+						// loop until we've seen the end newline
+						for seen+newlines < endln {
+							// Add more capacity (let append pick how much).
+							buf = append(buf, 0)[:len(buf)]
```
vikmik, line 363: This seems a bit inefficient given the initial buf size of 1024 bytes. A simple test like `TestAppenderRetryDocument/gzip` needs to grow this buffer 20 times

Without knowing more, I'd suggest using much more than 1024 bytes as initial capacity; maybe 16k? - we have so few bulkIndexers that it shouldn't make a material difference

vikmik:
Sorry I'm late to the party, but has it been considered to just track the document boundaries in a separate array? By recording the value of `buf.Len()` ) each time we add a doc. It looks like we could get away with 2 bytes per document to track that, which would avoid all this extra processing.
This seems this could save both on CPU (no need to decompress/recompress) and memory usage (2 bytes per document seems much smaller than having to use `copyBuf`, since we wouldn't need `copyBuf`, and wouldn't need to decompress+recompress. 

The only potential downside I can see is that we'd need to flush the gzip writer every time we add a document (se we can accurately record `buf.Len()` - but intuitively it seems the simplification and lowered memory usage seems worth it?

I might be missing something though (apologies if that's the case!)

axw:
> The only potential downside I can see is that we'd need to flush the gzip writer every time we add a document (so we can accurately record buf.Len() - but intuitively it seems the simplification and lowered memory usage seems worth it?

Compression of multiple documents would suffer though, so we would need to measure it. I think we should optimise for the happy path, and retries can be a little more expensive.

We could record the positions of the uncompressed documents though. We would still need to decompress/recompress, but we wouldn't need to scan for newlines.

vikmik:
Yeah, maybe that wasn't the best idea - what really prompted my reaction is that I am generally wary of optimizing for the happy path at the expense of the slow path when the slow path is more likely during operational events. In this case if ES is sending us 429s it probably means we're under higher load, and making things even more expensive when that happens seems like a slippery slope. In cases like this consistent performance is often preferable to peak performance.

PR #124, opened 2024-02-08
perf: try harder to operate on new data

kruskall:
With gzip the buffer can grow a lot, leading to a lot more work when
counting newlines. To improve performance we discard everything up to the
previous document when preparing the buffer for the next document.

Depends on https://github.com/elastic/go-docappender/pull/123


bulk_indexer.go
```diff
@@ -357,22 +357,40 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
+						endIdx += offset
 					}
 
 					b.writer.Write(buf[startIdx:endIdx])
+
+					if endIdx != len(buf) {
+						// the next document will be after the end of the current document
+						// so we can discard everything up to endIdx
+						seen += bytes.Count(buf[:endIdx], []byte{'\n'})
+						buf = buf[endIdx:]
```
vikmik, line 392: I found it fairly difficult to follow with these changes and the extra tracking. I've tried to simplify with a fixed-size buffer - see PR #129 - let me know what you think :)

vikmik (commented):
I still prefer what I did in #129 for the 2 reasons listed in the comments below. It is also less code than before, if we ignore the extra error handling I added

Is there anything you would change about #129? afaics it is safer, more performant and more concise. I didn't mean to step on your toes with this - apologies if it feels that way! I mostly did it because I'm already knee-deep in this code at the moment and I needed to understand the code closely :) I'd be happy to merge it if it works with you.

vikmik:
> > Is there anything you would change about #129? afaics it is safer, more performant and more concise. I didn't mean to step on your toes with this - apologies if it feels that way! I mostly did it because I'm already knee-deep in this code at the moment and I needed to understand the code closely :) I'd be happy to merge it if it works with you.
> 
> No worries 😄 I don't see any problem with #129 but I think error handling should be added separately from buffer reuse. Some of the changes also make the code harder to follow IMO, e.g. the `continue` statement of the change of the loop statement to `len(buff) == 0`

Sounds good - question of style choices - I find `continue` more readable as it reduces the context relevant to code below. But it's good either way :)
And you will laugh but I had added `len(buff) == 0` for readability only, because the code doesn't quite behave as the comments say otherwise. Now that our implementations are similar, your code has the same problem too - I'll explain it in a separate comment.

bulk_indexer.go
```diff
@@ -336,12 +336,7 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
-					// First loop, read from the gzip reader
-					if len(buf) == 0 {
-						n, _ := gr.Read(buf[:cap(buf)])
-						buf = buf[:n]
-					}
-
+					// newlines in the current buf
 					newlines := bytes.Count(buf, []byte{'\n'})
 
 					// loop until we've seen the start newline
```
vikmik, line 342: Now the for loop below ("loop until we've seen the start newline") isn't executed if the first document needs to be read (because `startln` == 0), which means the logic isn't very straightforward as it deviates from what the comments say. The code might still work but it feels like an accident. That's the reason why I left the extra `len(buf) == 0` in the for loop in #129. It works without, but understanding why the code works without that is quite confusing.

vikmik (approved):

PR #129, opened 2024-02-21
go-docappender: use a fixed-size buffer

vikmik:
This is an alternative implementation of PR #124 that doesn't require expanding the buffer, and requires fewer variables to track positions (hopefully improving maintainability)

### DIFF
--- a/bulk_indexer.go
+++ b/bulk_indexer.go
@@ -232,9 +232,10 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
 	if b.maxDocumentRetry > 0 {
 		if cap(b.copyBuf) < b.buf.Len() {
 			b.copyBuf = slices.Grow(b.copyBuf, b.buf.Len()-cap(b.copyBuf))
-			b.copyBuf = b.copyBuf[:cap(b.copyBuf)]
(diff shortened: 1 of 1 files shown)


bulk_indexer.go
```diff
@@ -232,9 +232,10 @@ func (b *bulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
 	if b.maxDocumentRetry > 0 {
 		if cap(b.copyBuf) < b.buf.Len() {
 			b.copyBuf = slices.Grow(b.copyBuf, b.buf.Len()-cap(b.copyBuf))
-			b.copyBuf = b.copyBuf[:cap(b.copyBuf)]
+			b.copyBuf = b.copyBuf[:b.buf.Len()]
 		}
-		copy(b.copyBuf, b.buf.Bytes())
+		n := copy(b.copyBuf, b.buf.Bytes())
+		b.copyBuf = b.copyBuf[:n]
```
vikmik, line 238: Prior to this PR there was no error handling of the gzip reader.
I fixed that below, but that required resizing down `copyBuf` here when necessary - otherwise there might be unrelated data from previous flushes causing failed reads.

vikmik:
Closing in favor of #124

PR #146, opened 2024-04-09
Accept elasticsearch clients other than v8

carsonip:
Make New* accept esapi.Transport instead of v8 implementation. This enables the use of v8 api on any Transport, e.g. v7 client in addition to existing v8 client. This should be safe enough for our very limited use case of bulk request with e.g. v7 client.

Add integration tests for v7.


carsonip:
> issue: the indexer still depends on the v8 package under the hood so passing a different version of the client might fail.

@kruskall if we are worried about v8 api not being compatible with v7 stack, which is possible if unsupported args are passed, WDYT about using v7 esapi to send to the backend, which can be v7, v8 or a future v9? Alternatively, go-docappender can have different code paths for v7 (use v7 esapi and v7 client) and v8 (use v8 esapi and v8 client), which might be an overkill given how minimal the difference is between v7 and v8 esapi. Either way on apm-server it will depend on both v7 and v8 packages.

What's your suggestion?

lahsivjar:
The changes overall look good to me. While it is definitely possible to break the compatibility between 2 esapi's and clients, I think, we can leave accepting the risk to the user of go-docappender. If we want, we can put a disclaimer(?) saying using any other client might work but is not officially supported.

elasticsearch.go
```diff
@@ -0,0 +1,24 @@
...
+// specific language governing permissions and limitations
+// under the License.
+
+package docappender
+
+import "net/http"
+
+type Transport interface {
+	Perform(*http.Request) (*http.Response, error)
+}
```
axw, line 24: Any particular reason we are not using `esapi.Transport` directly?

carsonip: good catch. Fixed.

lahsivjar (dismissed):
LGTM!

axw (dismissed):

carsonip:
Updated PR to address comment in https://github.com/elastic/go-docappender/pull/154#issuecomment-2060349472

go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/elastic/go-docappender
 go 1.21
 
 require (
+	github.com/elastic/go-elasticsearch/v7 v7.17.10
```
lahsivjar, line 6: WDYT about not having this dependency? I think if we separate the testing into a separate `go.mod` then we can do without this.

appender_integration_test.go
```diff
@@ -29,59 +29,110 @@ import (
...
+		err = indexer.Add(context.Background(), index, bytes.NewReader(encoded))
+		require.NoError(t, err)
+	}
+
+	// Closing the indexer flushes enqueued events.
+	err := indexer.Close(context.Background())
+	require.NoError(t, err)
+}
+
+func TestAppenderIntegrationV8(t *testing.T) {
```
lahsivjar, line 55: I don't think the testing is enough for the claim that we have tested with `v7`. I would suggest adding a wrapper for the tests that allow running ALL tests with both `v7` and `v8`. This would mean running tests for metric collection and other features that docappender has.

carsonip: Make sense. I can remove the comment about v7 support.

lahsivjar (approved):

PR #150, opened 2024-04-11
fix: update elastic.events.queued with the correct number of queued events

kruskall:
Because of document retries the number of queued events after a flush request can be >0 if the returned 429.
To fix the issue we only remove (failed+indexed) from the queued event metric after a flush request.

Closes https://github.com/elastic/go-docappender/pull/150

### DIFF
--- a/appender.go
+++ b/appender.go
@@ -374,7 +374,7 @@ func (a *Appender) flush(ctx context.Context, bulkIndexer *BulkIndexer) error {
 	}
 	docsFailed = int64(len(resp.FailedDocs))
 	totalFlushed := docsFailed + docsIndexed
-	defer a.addUpDownCount(-totalFlushed, &a.docsActive, a.metrics.docsActive)
+	a.addUpDownCount(-totalFlushed, &a.docsActive, a.metrics.docsActive)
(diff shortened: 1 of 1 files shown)


marclop (commented):
Thanks for the fix. Can you please add a test to verify this works as expected?

marclop:
I'm not convinced about pushing out tests for a bugfix. What's non-trivial about the test that needs to be added?

simitt:
@kruskall could you follow up on the last comment regarding the complexity of adding a test? I'm keen on getting this in, but right now the PR is stalling without additional tests.

appender.go
```diff
@@ -341,6 +340,7 @@ func (a *Appender) flush(ctx context.Context, bulkIndexer *BulkIndexer) error {
 		a.addCount(int64(flushed), &a.bytesUncompressedTotal, a.metrics.bytesUncompressedTotal)
 	}
 	if err != nil {
+		defer a.addUpDownCount(-int64(n), &a.docsActive, a.metrics.docsActive)
```
vikmik, line 343: Do we need the `defer` now? The actual flushing is done at this point. Same below

vikmik (approved):

Issue #127, opened 2024-02-19
Revisit buffer size when reading compressed content


kruskall:
Followup from https://github.com/elastic/go-docappender/pull/123#discussion_r1490075683

> This seems a bit inefficient given the initial buf size of 1024 bytes. A simple test like TestAppenderRetryDocument/gzip needs to grow this buffer 20 times
> 
> Without knowing more, I'd suggest using much more than 1024 bytes as initial capacity; maybe 16k? - we have so few bulkIndexers that it shouldn't make a material difference
> 

During retries we create a buf to temporarily store the (streamed) decompressed payload: https://github.com/elastic/go-docappender/blob/31a3978ba8f93c347ddbf193a9a94131d07f8952/bulk_indexer.go#L279

the buffer needs to be large enough to store 1 event.

As mentioned in the PR the current size is 1024 but should be increased to avoid growing the buffer when retrying.
The change is fairly small and doesn't change any logic. The increased memory usage should be negligible.

PR #153, opened 2024-04-16
feat: increase retry temp buffer size to 4096

kruskall:
Closes https://github.com/elastic/go-docappender/issues/127

Increate retry temp buffer size to 4096. Prevents growing the buffer multiple times


inge4pres (approved):
I can see now why we want do that 😄

PR #154, opened 2024-04-16
Accept elasticsearch clients other than v8, vendor a stripped down esapi

carsonip:
Alternative to #146

Do not depend on go-elasticsearch v8 esapi. Vendor a stripped down esapi to ensure v7 and v8 compatibility specifically for bulk request.

Both v7 and v8 are tested with integration tests.

Unfortunately v7 and v8 deps are in go.mod because of tests.


axw:
I'm not convinced this is better than https://github.com/elastic/go-docappender/pull/146. I'd be more inclined to keep the other PR and add integration tests that show it works with both v7 and v8, when you use the v7 transport.

carsonip:
Closing in favor of #146

PR #155, opened 2024-04-16
Change NewBulkIndexer signature to accept BulkIndexerConfig with a new Pipeline option; Release v2

carsonip:
Changes NewBulkIndexer interface to accept a config. The config includes a new Pipeline option, which will then be included in the bulk request.

Release v2 as NewBulkIndexer signature changed.


appender.go
```diff
@@ -141,7 +141,11 @@ func New(client *elasticsearch.Client, cfg Config) (*Appender, error) {
 	}
 	available := make(chan *BulkIndexer, cfg.MaxRequests)
 	for i := 0; i < cfg.MaxRequests; i++ {
-		available <- NewBulkIndexer(client, cfg.CompressionLevel, cfg.MaxDocumentRetries)
+		available <- NewBulkIndexer(client, BulkIndexerConfig{
```
marclop, line 144: To avoid breaking the API in the future, wouldn't it be better to also pass the Elasticsearch Client in `BulkIndexerConfig`?

carsonip: Agreed. I did this only for consistency with `func New(client esapi.Transport, cfg Config) (*Appender, error) {`. Will update this PR after #146

lahsivjar (dismissed):
Code LGTM!

> Although it breaks the NewBulkIndexer interface, I don't think this warrants a v2 as I do not expect any usage of NewBulkIndexer in the wild since BulkIndexer became public

Regarding this, I suggest we should stick to the convention and release v2.

bulk_indexer.go
```diff
@@ -138,14 +157,13 @@ func init() {
 
 // NewBulkIndexer returns a bulk indexer that issues bulk requests to Elasticsearch.
 // It is only tested with v8 go-elasticsearch client. Use other clients at your own risk.
-func NewBulkIndexer(client esapi.Transport, compressionLevel int, maxDocRetry int) *BulkIndexer {
+func NewBulkIndexer(cfg BulkIndexerConfig) *BulkIndexer {
```
marclop, line 160: It just occured to me that we're not validating the config options anywhere. If `Client` is nil, it will panic. I think this was okay before, since the type was unexported, but since the type is not exported and we're "properly" supporting it, it'd be worthy to add config validation and perhaps returning an `error` in the constructor.

Previously, the options were validated by the Appender's config, but now this type may be used standalone. WDYT?

carsonip: good point. I've added nil client check to appender as well.

go.mod
```diff
@@ -1,4 +1,4 @@
-module github.com/elastic/go-docappender
+module github.com/elastic/go-docappender/v2
```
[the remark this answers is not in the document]
carsonip: https://github.com/elastic/go-docappender/releases/tag/v1.1.0
We changed NewBulkIndexer signature compared to 1.1.0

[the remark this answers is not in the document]
carsonip: I might be missing something, but how is the `NewBulkIndexer` method not public in 1.1.0? It was only not public pre-1.1.0

Issue #136, opened 2024-03-13
publish metrics for retried documents


kruskall:
We're counting the number of retried docs:

https://github.com/elastic/go-docappender/blob/31a3978ba8f93c347ddbf193a9a94131d07f8952/bulk_indexer.go#L66

but we don't publish any metrics about it.

We should add a metric for "retried documents"  similar to what we are doing for Indexed, TooMany, etc.

The counter is already in the response stat and we should be able to publish the metric in https://github.com/elastic/go-docappender/blob/31a3978ba8f93c347ddbf193a9a94131d07f8952/appender.go#L284

simitt:
In addition to the above, the scope of this task should also contain to add the published metrics to our dashboards. Find a good place (potentially the index-service dashboard to add a visualization).

inge4pres:
Is `RetriedDocs` a counter that can be used also to infer how much times documents are retried?
Suppose we're setting `MaxRetry` to 10, is it possible to tell what fraction of those 10 retries are needed for a document to successfully being ingested?

PR #157, opened 2024-04-18
feat: add document retries metrics

kruskall:
Add a metric which tracks the number of document retries. We were already counting the retries but we never published metrics about it.
Update retry test to assert metrics count

Name is `elasticsearch.events.retried` to be consistent with other metrics. Feedback welcomed.

Closes https://github.com/elastic/go-docappender/issues/136


metric.go
```diff
@@ -103,6 +104,11 @@ func newMetrics(cfg Config) (metrics, error) {
 			description: "Number of APM Events flushed to Elasticsearch. Attributes are used to report separate counts for different outcomes - success, client failure, etc.",
 			p:           &ms.docsIndexed,
 		},
+		{
+			name:        "elasticsearch.events.retried",
+			description: "The number of document retries.",
```
carsonip, line 109:
```suggestion
			description: "The number of document retries. A single document may be retried more than once.",
```

endorama (approved):
Looks good to me!

PR #160, opened 2024-04-22
Fix allocation the document retry buffer

lahsivjar:
There were 2 bugs in the allocation of the `copyBuf` which is used to move the data around when document level retries are enabled:

1. The scaling of the length of `copyBuf` to its capacity was not done when the `copyBuf` didn't need to grow.
2. The logic used to calculate the count for `Grow` was incorrect. `Grow` ensures that the slice has room to grow by the count passed irrespective of its capacity. So if we grow a slice of capacity 10 and length 5 by 5 then there would be no changes required. The previous logic used `b.buf.Len() - cap(b.copyBuf)` to grow `copyBuf` to accommodate `buf` however, this would be incorrect, for example: if `buf.Len() = 1030` and `cap(b.copyBuf) = 1024` but `len(b.copyBuf) = 10` then calling `slices.Grow(b.copyBuf, b.buf.Len()-cap(b.copyBuf))` i.e. `slices.Grow(b.copyBuf, 6)` will not do anything as `b.copyBuf` can already append 6 elements.


bulk_indexer_test.go
```diff
@@ -0,0 +1,100 @@
...
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/elastic/go-docappender/v2"
+	"github.com/elastic/go-docappender/v2/docappendertest"
+	"github.com/stretchr/testify/require"
+)
+
+func TestBulkIndexer(t *testing.T) {
```
lahsivjar, line 33: [**For reviewers**] I wanted to write a black box test for the bulk indexer with large data (the current bug would only be detected if the document count was high). This led me to write this test which is slightly different than the appended test. Note that we don't currently test when compression is enabled -- this should also be done and I will create a followup for that later.

bulk_indexer.go
```diff
@@ -274,9 +274,9 @@ func (b *BulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
 
 	if b.config.MaxDocumentRetries > 0 {
 		if cap(b.copyBuf) < b.buf.Len() {
```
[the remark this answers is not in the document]
lahsivjar: No, it will panic if for some reason `cap(b.copyBuf) > b.buf.Len()` and while _I think_ it is not possible but I think being cautious here is better.

bulk_indexer.go
```diff
@@ -274,9 +274,9 @@ func (b *BulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
-			b.copyBuf = slices.Grow(b.copyBuf, b.buf.Len()-cap(b.copyBuf))
-			b.copyBuf = b.copyBuf[:cap(b.copyBuf)]
+			b.copyBuf = slices.Grow(b.copyBuf, b.buf.Len()-len(b.copyBuf))
 		}
+		b.copyBuf = b.copyBuf[:cap(b.copyBuf)]
 		n := copy(b.copyBuf, b.buf.Bytes())
 		b.copyBuf = b.copyBuf[:n]
```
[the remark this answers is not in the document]
lahsivjar: Grow does not mean that we only grow the capacity by the given number, it rather ensures we can *at-least* add n elements. So, I don't think we can simplify here.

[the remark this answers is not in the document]
lahsivjar: > 		b.copyBuf = b.copyBuf[:len(b.copyBuf)]

Also, this will not work since Grow doesn't change length, and copy works on length. We would have to use `b.buf.Len()`. I don't see a way we can simplify (as in reduce complexity).

carsonip (approved):
+1 on all your review comments. Thanks!

bulk_indexer.go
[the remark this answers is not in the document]
lahsivjar: I updated the code a bit to do only one scaling using the buffer length. Let me know if it meets your expectations.

carsonip (dismissed):
thanks, this is much more readable.

bulk_indexer_test.go
simitt: Appreciate the additional test here!

lahsivjar: @simitt Are you referring to adding more tests (with compression) in this PR itself?

lahsivjar: Added tests with different compression levels.

simitt: I meant to say that I appreciate that you have added the _TestBulkIndexer_ test.

carsonip (dismissed):

endorama (commented):
Thank you for the code clarity improvements! A nitpick but would be great adding some documentation on `buf` and `copyBuf`. The code is a bit cryptic on what is supposed to do with them and why.

PR #161, opened 2024-04-23
feat: add uncompressed flushed bytes metric

kyungeunni:
Currently, we collect request bytes flushed to Elasticsearch.

This PR introduces a metric that tracks the corresponding bytes before compression.


appender.go
```diff
@@ -60,18 +60,19 @@ var (
...
+	bulkRequests           int64
+	docsAdded              int64
+	docsActive             int64
+	docsFailed             int64
+	docsFailedClient       int64
+	docsFailedServer       int64
+	docsIndexed            int64
+	tooManyRequests        int64
+	bytesTotal             int64
+	bytesUncompressedTotal int64
```
kyungeunni, line 72: Due to the verbose/long name introduced, the indentation has changed and the readability of diff is not so great. I tried to shorten the name to 
- `byteUncompTotal`, but it was unnatural to read,
- `bytesRawTotal` was introducing cognitive load that raw and uncompressed bytes refer to the same thing

In the end, I chose `bytesUncompressedTotal`, but any suggestions are more than welcome.

metric.go
```diff
@@ -115,6 +116,12 @@ func newMetrics(cfg Config) (metrics, error) {
 			unit:        "by",
 			p:           &ms.bytesTotal,
 		},
+		{
+			name:        "elasticsearch.flushed.uncompressed.bytes",
```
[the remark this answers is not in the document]
kyungeunni: We recently added a metric named `server.request.uncompressed.bytes` in ingest service with https://github.com/elastic/apm-ingest-service/blob/main/internal/server/request/middleware/metric.go#L44.

For consistency, I think this also needs to be `uncompressed.bytes`, WDYT?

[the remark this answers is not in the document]
endorama: `uncompressed.bytes` looks consistent with  `elasticsearch.flushed.btyes` (see https://github.com/elastic/go-docappender/pull/161/files#diff-1fe046256e4fc97d49ef96e6b960b5dddf4198c29861d6ae24a8c97771c04010R191) metric, I would favor that unless there are other compelling reasons not to.

bulk_indexer.go
```diff
@@ -218,11 +233,21 @@ func (b *BulkIndexer) Len() int {
...
+	return b.writer.count
+}
+
 // BytesFlushed returns the number of bytes flushed by the bulk indexer.
 func (b *BulkIndexer) BytesFlushed() int {
 	return b.bytesFlushed
 }
 
+// BytesUncompFlushed returns the number of uncompressed bytes flushed by the bulk indexer.
+func (b *BulkIndexer) BytesUncompFlushed() int {
```
endorama, line 247: nit: I'm fine with using longer names where it helps readability and I think would be good for consistency to use `BytesUncompressedFlushed()` here.

endorama (approved):
A nitpick but overall looks good to me!

bulk_indexer.go
```diff
@@ -161,6 +162,16 @@ func init() {
 	})
 }
 
+type countWriter struct {
+	count int
```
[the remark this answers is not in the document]
kyungeunni: bytesWritten is more clear and descriptive, updated  👍

docappendertest/docappendertest.go
```diff
@@ -44,7 +45,10 @@ import (
 const TimestampFormat = "2006-01-02T15:04:05.000Z07:00"
 
 // DecodeBulkRequest decodes a /_bulk request's body, returning the decoded documents and a response body.
-func DecodeBulkRequest(r *http.Request) ([][]byte, esutil.BulkIndexerResponse) {
+func DecodeBulkRequest(r *http.Request) (
```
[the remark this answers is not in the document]
kyungeunni: > move countReader out of here and wrap the r.Body before passing it to DecodeBulkRequest. We can directly count the size in the caller rather than here.

If we wrapped from the caller, the count would be the compressed bytes read by the gzip reader in case the compression was enabled so we won't able to track the uncompressed bytes. I ended up adding a new method to return the RequestStats struct.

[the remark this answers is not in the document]
lahsivjar: Ah good point!

appender.go
```diff
@@ -333,6 +335,11 @@ func (a *Appender) flush(ctx context.Context, bulkIndexer *BulkIndexer) error {
 	if flushed := bulkIndexer.BytesFlushed(); flushed > 0 {
 		a.addCount(int64(flushed), &a.bytesTotal, a.metrics.bytesTotal)
 	}
+	// Record the BulkIndexer buffer's length as the bytesTotal metric after
```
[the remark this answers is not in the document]
kyungeunni: Thanks, updated 👍

bulk_indexer.go
```diff
@@ -396,7 +423,7 @@ func (b *BulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
 					if len(buf) == 0 {
 						n, err := gr.Read(buf[:cap(buf)])
 						if err != nil && err != io.EOF {
-							return resp, fmt.Errorf("failed to read from compressed buffer: %w", err)
+							return resp, fmt.Errorf("failed to read from uncompressed buffer: %w", err)
```
lahsivjar, line 426: I am a bit confused here, I think `compressed` is more correct? What was your reasoning to use `uncompressed`?

docappendertest/docappendertest.go
```diff
@@ -85,6 +86,56 @@ func DecodeBulkRequest(r *http.Request) ([][]byte, esutil.BulkIndexerResponse) {
 	return indexed, result
 }
 
+// DecodeBulkRequestWithStats decodes a /_bulk request's body, returning the decoded documents
+// and a response body and stats about request.
+func DecodeBulkRequestWithStats(r *http.Request) (
```
lahsivjar, line 91: I suggest we use `DecodeBulkRequestWithStats` in `DecodeBulkRequest` and just discard the stats. This will avoid duplicating the code.

bulk_indexer.go
kyungeunni: Oops, I was mistaken here, it reads from the compressed buffer and we get uncompressed data. will revert the change.

docappendertest/docappendertest.go
kyungeunni: Makes sense, addressed 👍

lahsivjar (approved):

bulk_indexer.go
kyungeunni: I agree! updated.

PR #165, opened 2024-05-02
feat: optionally enable logging the full error reason

kyungeunni:
Log the full error reason when configured.

This shouldn't be enabled by default as it is advised to redact the full context.


axw:
@kyungeunni please make sure the redaction happens in BulkIndexer, before returning any results back to Appender

axw (approved):
looks great, thanks!

PR #167, opened 2024-05-03
feat: capture caused_by and return it as part of `BulkIndexerResponseItem`

### LINKED ISSUE #166: Capture "caused_by" in errors
If one attempts to index a very long value in a `flattened` field, it will cause indexing to fail. This is reported by `go-docappender`, but the reason for the error is obscured. It will look something like `"reason":"[1:40656] failed to parse field [error.exception.stacktrace] of type [flattened] in document with id '5UtbPI8BaMhqs2kExrLc'.`. Elasticsearch responds with a preview of the field value, but we strip this off.

The Elasticsearch response would also include "caused_by", which has helpful information about what caused the indexing to fail. In the example scenario above, it would be something like `"caused_by":{"type":"illegal_argument_exception","reason":"Flattened field [error.exception.stacktrace] contains one immense field whose keyed encoding is longer than the allowed max length of 32766 bytes. Key length: 17, value length: 121034 for key starting with [vars.request_body]"}`.

We should consider logging the cause to assist in debugging.

kyungeunni:
closes #166 

Parse `caused_by` from the response, and add it to `BulkIndexerResponseItem` so that it can be logged.

#### TODO
- [ ] Decide whether to make it optional (verbose:on/off)
- [ ] Manual testing

### DIFF
--- a/appender.go
+++ b/appender.go
@@ -391,9 +391,10 @@ func (a *Appender) flush(ctx context.Context, bulkIndexer *BulkIndexer) error {
 		}
 	}
 	for key, count := range failedCount {
-		logger.Error(fmt.Sprintf("failed to index documents in '%s' (%s): %s",
+		logger.Error(fmt.Sprintf("failed to index documents in '%s' (%s): %s, caused_by (%s): %s",
 			key.Index, key.Error.Type, key.Error.Reason,
-		), zap.Int("documents", count))
+			key.Error.Cause.Type, key.Error.Cause.Reason),
+			zap.Int("documents", count))
 	}
 	if docsFailed > 0 {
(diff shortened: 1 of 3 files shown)


axw:
>  Decide whether to make it optional (verbose:on/off)

This bit is critical. Do you have a plan for how we are going decide this?
I think the PR should probably be in draft until this is decided.

kyungeunni:
@axw Having discussed this with the team, it seems it'd make sense to log this by default as it would be helpful to understand the error better even though the logs become a bit verbose. Do you have any concern about this?

axw:
I'm not concerned about verbosity. My only concern is whether we might leak document details in the `caused_by`. Do we know that that won't happen?

PR #168, opened 2024-05-14
metric: add success dimension to `elasticsearch.flushed.uncompressed.bytes`

kyungeunni:
Should be ready to review when knowing the increased memory usage


kyungeunni:
Closing this as we are not going to go ahead with this solution after observing the benchmark result

PR #180, opened 2024-06-07
Support `require_data_stream` new config option

### LINKED ISSUE #110: Add config to add `require_data_stream` to indexing requests
https://github.com/elastic/elasticsearch/pull/101872 has been merged and will be part of 8.13.0. Let's update go-docappender to optionally set this flag in indexing requests.

marclop:
Adds a new `RequireDataStream` option to the Appender, and BulkIndexer configurations. When set, it will send `?require_data_stream=true` in the `_bulk` request to Elasticsearch.

Closes #110

### DIFF
--- a/config.go
+++ b/config.go
@@ -90,10 +90,9 @@ type Config struct {
 	// If Pipeline is empty, no ingest pipeline will be specified in the Bulk request.
 	Pipeline string
 
-	// RequireDataStream forces the auto creation of the underlying index.
-	// If set to true, an index will be created only if a matching index
-	// template is found and it contains a data stream template. When true,
-	// `require_data_stream=true` is set in the bulk request.
+	// RequireDataStream, If set to true, an index will be created only if a
+	// matching index template is found and it contains a data stream template.
+	// When true, `require_data_stream=true` is set in the bulk request.
 	//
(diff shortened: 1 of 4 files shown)


config.go
```diff
@@ -90,6 +90,14 @@ type Config struct {
 	// If Pipeline is empty, no ingest pipeline will be specified in the Bulk request.
 	Pipeline string
 
+	// RequireDataStream forces the auto creation of the underlying index.
```
carsonip, line 93: This line doesn't read very right, although the explanation below is correct.

marclop: you're right, just re-formatted it.

endorama (dismissed):
Looks good to me!

config.go
```diff
@@ -90,6 +90,13 @@ type Config struct {
 	// If Pipeline is empty, no ingest pipeline will be specified in the Bulk request.
 	Pipeline string
 
+	// RequireDataStream, If set to true, an index will be created only if a
```
endorama, line 93: A true nitpick but could be beneficial to explain the default behavior and why one may want enable this option, to help readers unfamiliar with it.

marclop: Documented the default behavior commit below.

carsonip (approved):

Issue #45, opened 2023-07-13
Add support for OTel tracing


marclop:
## Description

The appender currently accepts an `*apm.Tracer` in its config and uses it to trace flushes to Elasticsearch. 

https://github.com/elastic/go-docappender/blob/96531c337391cc64310d9ba6562a2b39f214015e/config.go#L38-L42

To align our general efforts, we should consider using OTel tracing, which also aligns with #37.

axw:
I apparently missed the news that go-elasticsearch has OTel tracing now. We should probably be relying on that: https://github.com/elastic/go-elasticsearch/blob/264743bde3a913cb8aa84160fcef7d5551c44696/elasticsearch.go#L117-L132

endorama:
Is this task related to adding support to Otel in parallel with current tracing or should the current tracer be removed entirely? In the latter, it looks like a breaking change to me. I consider part of this task a release including the new tracer, do you agree on the breaking change?

axw:
@endorama yes, it would be breaking. I think we should add (opt-in) OTel tracing in parallel with the existing Elastic APM tracing, and later follow up with removing Elastic APM instrumentation.

endorama:
PR for supporting otel tracing is ready for review in https://github.com/elastic/go-docappender/pull/182

Following up on this @axw comment
> I apparently missed the news that go-elasticsearch has OTel tracing now. We should probably be relying on that: https://github.com/elastic/go-elasticsearch/blob/264743bde3a913cb8aa84160fcef7d5551c44696/elasticsearch.go#L117-L132

To my understanding in `go-docappender` we receive an already instantiated client (in `New(client esapi.Transport, ...)`. Due to this is not `go-docappender` responsibility to leverage `go-elasticsearch` instrumentation but it's caller responsibility to provide a client with instrumentation initialized. 
The changes in the linked PR are to add the span from `Appender.flush`, which would otherwise be lost, but do not include tracing Elasticsearch client, as that comes from the client.

endorama:
I created https://github.com/elastic/go-docappender/issues/184 to follow up on removing the APM tracer marked as deprecated.

PR #182, opened 2024-06-12
add otel tracing support

endorama:
Add otel tracing support as a secondary tracing choice.

By default the current behaviour is retained: tracing with Elastic APM is preferred over otel tracing.
If Elastic APM tracing is not configured but otel tracing is, leverage otel tracing.

If no tracing is configured keep the current behaviour of skipping tracing.

Closes https://github.com/elastic/go-docappender/issues/45

### DIFF
--- a/appender.go
+++ b/appender.go
@@ -327,7 +327,6 @@ func (a *Appender) flush(ctx context.Context, bulkIndexer *BulkIndexer) error {
 		// below with the trace.
 		logger = logger.With(apmzap.TraceContext(ctx)...)
 	} else if a.otelTracingEnabled() {
-		// NOTE: this is missing transaction type information. How is this conveyed in otel?
 		ctx, span = a.tracer.Start(ctx, "docappender.flush", trace.WithAttributes(
 			attribute.Int("documents", n),
 		))
--- a/appender.go
+++ b/appender.go
@@ -718,6 +718,8 @@ func (a *Appender) tracingEnabled() bool {
 	return a.config.Tracer != nil && a.config.Tracer.Recording()
 }
 
+// otelTracingEnabled checks whether we should be doing tracing
+// using otel tracer.
 func (a *Appender) otelTracingEnabled() bool {
 	return a.tracer != nil
 }
--- a/appender.go
+++ b/appender.go
@@ -33,6 +33,7 @@ import (
 	"go.elastic.co/apm/module/apmzap/v2"
 	"go.elastic.co/apm/v2"
 	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
 	"go.opentelemetry.io/otel/metric"
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
@@ -474,6 +475,9 @@ func (a *Appender) flush(ctx context.Context, bulkIndexer *BulkIndexer) error {
 		zap.Int64("docs_failed", docsFailed),
 		zap.Int64("docs_rate_limited", tooManyRequests),
 	)
+	if a.otelTracingEnabled() && span.IsRecording() {
+		span.SetStatus(codes.Ok, "")
+	}
 	return nil
 }
 
--- a/appender.go
+++ b/appender.go
@@ -88,6 +88,7 @@ type Appender struct {
 	mu                    sync.Mutex
 	closed                chan struct{}
 
+        // tracer is an OTel tracer, and should not be confused with `a.config.Tracer` which is an Elastic APM Tracer.
 	tracer trace.Tracer
 }
 
--- a/appender.go
+++ b/appender.go
@@ -692,7 +692,8 @@ func (a *Appender) indexFailureRate() float64 {
 
 // tracingEnabled checks whether we should be doing tracing
 func (a *Appender) tracingEnabled() bool {
-	return a.config.Tracer != nil && a.config.Tracer.Recording()
+	return (a.config.Tracer != nil && a.config.Tracer.Recording()) ||
+		a.config.OtelTracer != nil
 }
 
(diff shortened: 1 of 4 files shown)


appender.go
```diff
@@ -316,6 +318,19 @@ func (a *Appender) flush(ctx context.Context, bulkIndexer *BulkIndexer) error {
...
+			attribute.Int("documents", n),
+		))
+		defer span.End()
+
+		// Add trace IDs to logger, to associate any per-item errors
+		// below with the trace.
+		logger = logger.With(
+			zap.String("traceId", span.SpanContext().TraceID().String()),
+			zap.String("spanId", span.SpanContext().SpanID().String()),
+		)
```
endorama, line 333: For reviewers: A otel zap bridge exists at https://github.com/uptrace/opentelemetry-go-extra/tree/main/otelzap
This implementation is not compatible with a standard zap logger though, as it wraps the logger in a custom `otelzap.Logger` type. Passing the context also requires to use an additional `.Ctx(ctx)` call that is not supported by `zap.Logger`.
Due to this I chose not to use it an instead provide correlation information as standard otel attributes.

appender_test.go
```diff
@@ -1647,6 +1649,51 @@ func TestAppenderTracing(t *testing.T) {
 	testAppenderTracing(t, 400, "failure")
 }
 
+func TestAppenderOtelTracing(t *testing.T) {
```
endorama, line 1652: This test is currently failing, as reported spans & observed logs is `0`.

config.go
```diff
@@ -43,6 +44,8 @@ type Config struct {
 	// If Tracer is nil, requests will not be traced.
 	Tracer *apm.Tracer
 
+	OtelTracer trace.Tracer
```
endorama, line 47: This is missing some documentation that I'll provide before opening for review.

appender.go
```diff
@@ -397,6 +414,9 @@ func (a *Appender) flush(ctx context.Context, bulkIndexer *BulkIndexer) error {
 		failedCount[info]++
 		if a.tracingEnabled() {
 			apm.CaptureError(ctx, errors.New(info.Error.Reason)).Send()
+		} else if a.otelTracingEnabled() {
+			// trace.SpanFromContext(ctx).RecordError(errors.New(info.Error.Reason))
+			span.RecordError(errors.New(info.Error.Reason))
```
endorama, line 419: Is possible instead of using a `span` var in the local scope (initialized on line 311) to always extract it from the context. I don't see benefits in doing so but I've yet to finalize a decision on what to use.

appender_test.go
endorama: Test now fixed, I was shutting down the exporter, but being it in memory this was discarding the spans.

appender.go
endorama: Decided to use `span` variable.

1pkg (dismissed):

config.go
```diff
@@ -40,9 +41,18 @@ type Config struct {
...
+	// OtelTracerProvider may not be nil, in which case the request will
+	// be traced by a different tracer.
 	Tracer *apm.Tracer
 
+	// OtelTracerProvider holds an optional otel TracerProvider for tracing
+	// flush requests.
+	//
+	// If OtelTracerProvider is nil, requests will not be traced.
+	// To use this provider Tracer must be nil.
+	OtelTracerProvider trace.TracerProvider
```
axw, line 54: Maybe call this `TracerProvider`, with the expectation that we'll remove `Tracer` in the future.

config.go
```diff
@@ -40,9 +41,18 @@ type Config struct {
 	// Tracer holds an optional apm.Tracer to use for tracing bulk requests
 	// to Elasticsearch. Each bulk request is traced as a transaction.
 	//
-	// If Tracer is nil, requests will not be traced.
+	// If Tracer is nil, requests will not be traced. Note however that
+	// OtelTracerProvider may not be nil, in which case the request will
+	// be traced by a different tracer.
```
axw, line 46: I'm tempted to add a deprecation notice to this, WDYT?

axw (approved):

config.go
endorama: I agree, done in [`a5fa1ad` (#182)](https://github.com/elastic/go-docappender/pull/182/commits/a5fa1ad7198cf9373558453db310f8588327a5b8)

config.go
endorama: done in [`6cc6cea` (#182)](https://github.com/elastic/go-docappender/pull/182/commits/6cc6ceac993cb58d2e7e1fba7ad1ec070a3ca63f)

appender.go
```diff
@@ -85,6 +87,8 @@ type Appender struct {
 	metrics               metrics
 	mu                    sync.Mutex
 	closed                chan struct{}
+
+	tracer trace.Tracer
```
carsonip, line 91: a.config.tracer is an Elastic APM tracer but a.tracer is an otel tracer, might be confusing to one. Is it possible to make it clearer?

carsonip: ah, I see that it was discussed. So maybe add a comment that this should not be confused with a.config.Tracer?

carsonip, line 91:
```suggestion
        // tracer is an OTel tracer, and should not be confused with `a.config.Tracer` which is an Elastic APM Tracer.
	tracer trace.Tracer
```

endorama: Thanks for the suggestion, I approved it. Once the migration to otel is completed I think this will look way less confusing, but meanwhile a clarification comment helps for sure.

carsonip (approved):

carsonip:
## Testing notes

:question: test-plan-skip

This change should not concern apm-server, as apm-server explicitly passes cfg.Tracer which is an elastic apm tracer instead of an otel TracerProvider to docappender. Also, judging by the code, it should not have any side effect on existing use cases.

PR #189, opened 2024-07-24
Add dynamic templates to bulk indexer item

carsonip:
Make bulk request accept dynamic templates by adding DynamicTemplates field to bulk indexer item. This is only added to bulk indexer, not appender.


bulk_indexer.go
```diff
@@ -287,18 +288,39 @@ func (b *BulkIndexer) Add(item BulkIndexerItem) error {
...
 	if index != "" {
-		if documentID != "" {
+		if !first {
 			b.jsonw.RawByte(',')
 		}
 		b.jsonw.RawString(`"_index":`)
 		b.jsonw.String(index)
+		first = false
+	}
+	if dynamicTemplates != nil && len(dynamicTemplates) > 0 {
```
axw, line 307:
```suggestion
	if len(dynamicTemplates) > 0 {
```

axw (commented):
Looks good. Could we add a small unit test that validates the _bulk request format?

bulk_indexer.go
```diff
@@ -269,14 +269,15 @@ func (b *BulkIndexer) BytesUncompressedFlushed() int {
 }
 
 type BulkIndexerItem struct {
-	Index      string
-	DocumentID string
-	Body       io.WriterTo
+	Index            string
+	DocumentID       string
+	Body             io.WriterTo
+	DynamicTemplates map[string]string
```
carsonip, line 275: @axw I was initially a little hesitant to pass in a map here, as Body is a io.WriterTo containing serialized bytes which makes it look inconsistent. But now I'm fine with the idea, as DynamicTemplates does not really require specific data structures and is only a `map[string]string`. WDYT?

axw: I think of it like this:
 - `DynamicTemplates` has a known structure, which fits naturally to `map[string]string`
 - `Body` just needs to be JSON, but we shouldn't be too prescriptive about how the JSON encoding is done

So I think it's perfectly fine to use `map[string]string` here.

axw (approved):
LGTM!

PR #191, opened 2024-08-02
ci: enable race detector for go tests

kruskall:
`-race` was added for the make task in https://github.com/elastic/go-docappender/pull/67 but the ci workflow is still running plain go test: https://github.com/elastic/go-docappender/blob/068fe43e1e1913dbbd86a8a078e8fd3d9d6dde67/.github/workflows/ci.yml#L40


1pkg (approved):

marclop:
Another failed test, seems like it needs fixing

PR #205, opened 2024-10-08
fix: decode and format error response from es

kruskall:


bulk_indexer.go
```diff
@@ -389,7 +390,27 @@ func (b *BulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
+				Type     string `json:"type,omitempty"`
+				Reason   string `json:"reason,omitempty"`
+				CausedBy struct {
+					Type   string `json:"type,omitempty"`
+					Reason string `json:"reason,omitempty"`
+				} `json:"caused_by,omitempty"`
+			} `json:"error,omitempty"`
+		}
+
+		if err := jsoniter.NewDecoder(res.Body).Decode(&er); err == nil {
```
1pkg, line 405: if an error was returned here when decoding, should we consider fallback to `s = res.String()`?

1pkg (approved):

bulk_indexer.go
```diff
@@ -389,7 +390,27 @@ func (b *BulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
+			er.Error.Reason = ""
```
endorama, line 406: What do you think about adding a note here explaining why we are removing these fields? Looks something easy to overlook in a future refactoring.

endorama (approved):

PR #209, opened 2024-10-23
`elasticsearch.events.retried`: Add retry dimension

marclop:
This commit adds a new dimension to the `elasticsearch.events.retried` metric called `greatest_retry`. It represents the greatest observed retry count for a bulk request. This is useful to understand how many times documents in a bulk request have been retried.


inge4pres (approved):
Well done Marc, thanks 🙏🏼

PR #219, opened 2024-12-10
add metric: docappender.blocked.add counter

endorama:
This monotonic counter increments every time `Appender.Add`
detects that `bulkItems` channel length and capacity are
equal before sending the created BulkIndexerItem to the
channel.

As `bulkItems` is a buffered channel, monitoring length and
capacity allow us to detect possible blocking events when
sending to the channel due to exhausted capacity.

As `Appender.Add` could be called from multiple goroutines,
we are not guaranteed that every time this metric is
incremented the channel send will block, but should give us
a reasonably approximated picture.

### DIFF
--- a/appender.go
+++ b/appender.go
@@ -75,6 +75,7 @@ type Appender struct {
 	availableBulkRequests  int64
 	activeCreated          int64
 	activeDestroyed        int64
+	blockedAdd             int64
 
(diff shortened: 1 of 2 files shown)


inge4pres (approved):
🚀

PR #227, opened 2025-01-23
Add per event pipeline

leehinman:
Adds pipeline to the BulkIndexItem.  This will be necessary to implement https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/37419

### DIFF
--- a/bulk_indexer_test.go
+++ b/bulk_indexer_test.go
@@ -165,3 +165,41 @@ func TestDynamicTemplates(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, int64(2), stat.Indexed)
 }
+
+func TestPipeline(t *testing.T) {
+	client := docappendertest.NewMockElasticsearchClient(t, func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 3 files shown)


bulk_indexer_test.go
```diff
@@ -165,3 +165,36 @@ func TestDynamicTemplates(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, int64(2), stat.Indexed)
 }
+
+func TestPipeline(t *testing.T) {
+	client := docappendertest.NewMockElasticsearchClient(t, func(w http.ResponseWriter, r *http.Request) {
```
carsonip, line 170: would be good to assert that pipeline is successfully passed in the request

carsonip (commented):
thanks. an issue with the test, otherwise lgtm

mauri870 (approved):

bulk_indexer_test.go
leehinman: Updated the test so it checks that the pipelines made it.

carsonip (approved):

leehinman:
@carsonip Looks like reviews and CI checks are green.  I don't have write access to the repo so I can't merge, can you?

Issue #211, opened 2024-11-05
TestAppenderIndexFailedLogging is Flaky


rubvs:
```
=== RUN   TestAppenderIndexFailedLogging
    appender_test.go:797:
        	Error Trace:	/Users/rubenvanstaden/elastic/go-docappender/appender_test.go:797
        	Error:      	"[{{error 2024-11-07 17:06:51.161592 -0500 EST m=+0.306465626  failed to index documents in 'an_index' (document_parsing_exception):  undefined } [{documents 11 1  <nil>}]} {{error 2024-11-07 17:06:51.161538 -0500 EST m=+0.306411376  failed to index documents in 'an_index' (error_type): error_reason_even undefined } [{documents 11 1  <nil>}]} {{error 2024-11-07 17:06:51.161601 -0500 EST m=+0.306474751  failed to index documents in 'an_index' (error_type): error_reason_even undefined } [{documents 11 2  <nil>}]} {{error 2024-11-07 17:06:51.161568 -0500 EST m=+0.306440959  failed to index documents in 'an_index' (error_type): error_reason_odd undefined } [{documents 11 1  <nil>}]} {{error 2024-11-07 17:06:51.161612 -0500 EST m=+0.306485709  failed to index documents in 'an_index' (error_type): error_reason_odd undefined } [{documents 11 1  <nil>}]} {{error 2024-11-07 17:06:51.161576 -0500 EST m=+0.306449709  failed to index documents in 'an_index' (unavailable_shards_exception):  undefined } [{documents 11 1  <nil>}]} {{error 2024-11-07 17:06:51.161621 -0500 EST m=+0.306494751  failed to index documents in 'an_index' (unavailable_shards_exception):  undefined } [{documents 11 1  <nil>}]} {{error 2024-11-07 17:06:51.161579 -0500 EST m=+0.306452168  failed to index documents in 'an_index' (x_content_parse_exception):  undefined } [{documents 11 1  <nil>}]} {{error 2024-11-07 17:06:51.161585 -0500 EST m=+0.306458001  failed to index documents in 'an_index' (x_content_parse_exception):  undefined } [{documents 11 1  <nil>}]}]" should have 5 item(s), but has 9
        	Test:       	TestAppenderIndexFailedLogging
--- FAIL: TestAppenderIndexFailedLogging (0.00s)
FAIL
FAIL	github.com/elastic/go-docappender/v2	2.342s
FAIL
```

## Test

- Race detector has to be enabled
```
[...]

PR #228, opened 2025-01-29
Failure Store: Update go-docappender to respect failure store status

1pkg:
# Description

This PR updates go-docappender to respect new failure store response status and emit new correspodning metrics accordingly. The old "indexed" metrics stay intact instead a new separate set of failure store labels is exposed.

This PR depends on changes made for `BulkIndexerResponseItem` in https://github.com/elastic/go-elasticsearch/pull/948 and should only be merged afterwards. 

# How to test:

Use an instance of ES that has failure store feature enabled and enable failure store for a data stream via component template with.

```
POST _component_template/$name
{
  "template": {
    "data_stream_options": {
      "failure_store": {
        "enabled": true
      }
    }
  }
}
```

Set a custom "fail" ingest pipeline with.

```
PUT _ingest/pipeline/$name
{
  "processors": [
    {
      "fail": {
        "message": "fail"
      }
    }
  ]
}
```

Ingest some data to corresponding data stream, then check that failure store metrics are getting reported correctly.

To emulate "failed" status, set backing data stream index to be read only with.

```
PUT /$index/_settings
{
  "index.blocks.read_only": true
}
``` 

image


appender_test.go
```diff
@@ -753,7 +753,6 @@ func TestAppenderFlushRequestError(t *testing.T) {
 			}
```
1pkg, line 753: fix for a flaky test

appender_test.go
```diff
@@ -173,6 +179,9 @@ loop:
 			case "TooMany":
 				processedAsserted++
 				assert.Equal(t, stats.TooManyRequests, dp.Value)
+			case "FailureStore":
+				processedAsserted++
+				assert.Equal(t, stats.TooManyRequests, dp.Value)
```
axw, line 184: Is `stats.TooManyRequest` meant to be `stats.FailureStoreUsed/Failed/Something`?

axw (commented):
Looks good! Just one question

appender_test.go
1pkg: Good catch, a bad copy from the above.
Fixed this now.

axw:
One more question, sorry: should we be measuring "not_enabled" too?

1pkg:
> One more question, sorry: should we be measuring "not_enabled" too?

Agree, this metric could be useful too. Updated to expose it in the last commit.

go.mod
```diff
@@ -1,9 +1,9 @@
 module github.com/elastic/go-docappender/v2
 
-go 1.22
+go 1.22.0
 
 require (
-	github.com/elastic/go-elasticsearch/v8 v8.17.0
+	github.com/elastic/go-elasticsearch/v8 v8.4.0-alpha.1.0.20250131174650-1dda5df4f11f
```
1pkg, line 6: Note for reviewers: unfortunately we have to use the latest unreleased go-elasticsearch library that includes https://github.com/elastic/go-elasticsearch/pull/948 for now.

marclop: Are the patch versions also aligned with the stack? That seems strange.

appender.go
```diff
@@ -63,20 +63,23 @@ var (
 // server to make progress encoding while Elasticsearch is busy servicing flushed bulk requests.
 type Appender struct {
 	// legacy metrics for Stats()
```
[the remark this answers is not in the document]
1pkg: updated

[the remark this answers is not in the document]
1pkg: Want to double check, do we plan to remove `Stats` API completely in the future?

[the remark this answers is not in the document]
marclop: Yes, we'll remove Stats completely in the future.

marclop (commented):
I think this looks good overall, just one small question

bulk_indexer.go
```diff
@@ -114,6 +114,16 @@ type BulkIndexerResponseStat struct {
...
+	// FailureStore contains failure store specific stats.
+	FailureStore struct {
+		// Used contains the total number of documents indexed to failure store.
+		Used int64
+		// Failed contains the total number of documents which failed when indexed to failure store.
+		Failed int64
+		// NotEnabled contains the total number of documents which could have been indexed to failure store
+		// if it was enabled.
+		NotEnabled int64
+	}
```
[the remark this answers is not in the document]
1pkg: Correct me if I'm wrong, but per my understanding legacy stats metrics are residing in https://github.com/elastic/go-docappender/blob/main/appender.go#L65. Which I already reverted in https://github.com/elastic/go-docappender/pull/228/commits/48f9d850971041b813e6f9e637ef3cb22e4f009c. 

While this is just generic bulk response container which is used to pass data to appender for reporting OTEL metrics, etc.

appender.go
```diff
@@ -483,11 +483,41 @@ func (a *Appender) flush(ctx context.Context, bulkIndexer *BulkIndexer) error {
 			metric.WithAttributes(attribute.String("status", "FailedServer")),
 		)
 	}
+	if failureStore.Used > 0 {
+		a.addCount(failureStore.Used, nil,
+			a.metrics.docsIndexed,
+			metric.WithAttributes(
+				attribute.String("status", "FailureStore"),
```
simitt, line 490: I'm not sure how the `status: "FailureStore"` is supposed to be used?
E.g. when calculating SLOs and being interested in success vs failed documents. If I understand it correct then the `failure_store: used` documents would be counted as `success` whereas `failure_store: failed` and `failure_store: not_enabled` would be counted as `failed`?

bulk_indexer.go
```diff
@@ -134,6 +144,22 @@ type BulkIndexerResponseItem struct {
 	} `json:"error,omitempty"`
 }
 
+// FailureStoreStatus defines enumeration type for all known failure store statuses.
+type FailureStoreStatus string
+
+const (
+	// FailureStoreStatusUnknown implicit status which represents that there is no information about
+	// this response or that the failure store is not applicable.
+	FailureStoreStatusUnknown FailureStoreStatus = "not_applicable_or_unknown"
```
simitt, line 153: how about just `unknown`? 
Is this used somewhere?

bulk_indexer.go
```diff
@@ -114,6 +114,16 @@ type BulkIndexerResponseStat struct {
...
 	RetriedDocs int64
+	// FailureStore contains failure store specific stats.
+	FailureStore struct {
+		// Used contains the total number of documents indexed to failure store.
+		Used int64
+		// Failed contains the total number of documents which failed when indexed to failure store.
+		Failed int64
+		// NotEnabled contains the total number of documents which could have been indexed to failure store
+		// if it was enabled.
+		NotEnabled int64
```
simitt, line 125: The already reported stats are not very consistent (some of them contain the term `docs`, others don't - e.g. `Indexed` vs `RetriedDocs, FailedDocs`). 
However, when reading `FailureStore.Used`, `FailureStore.Failed` and `FailureStore.NotEnabled`, it's not necessarily clear from the naming whether this would count the number of documents or batch requests. Is it too verbose to add `Docs`?

bulk_indexer.go
1pkg: It's mapped directly from ES definition [in](https://github.com/elastic/elasticsearch/blob/1225b0720aba89247c14727bf2e8e29a1af96cd0/server/src/main/java/org/elasticsearch/action/bulk/IndexDocFailureStoreStatus.java#L42), we can remap it if we think it adds to readability.

bulk_indexer.go
1pkg: I see no problem adding docs to the names. What makes more sense `FailureStoreDocs.Used`, `FailureStoreDocs.Failred`, `FailureStoreDocs.NotEnabled` or alternatively `FailureStore.UsedDocs`, `FailureStore.FailredDocs`, `FailureStore.NotEnabledDocs`?

appender.go
1pkg: My idea behind it that the existing SLOs stay untouched, with enabled failure store they should always be 100% good events in theory. Separately a new set of SLOs can be created to track failure store error rate, and the number of documents indexed to FS vs all documents.

simitt: sounds good, thanks

bulk_indexer.go
simitt: ah, hadn't made this connection - let's keep it as is then

bulk_indexer.go
simitt: I'm slightly in favour of `FailureStoreDocs.Used`, but no strong opinion.

1pkg: Updated in the last commit

marclop (approved):
LGTM!

appender_test.go
```diff
@@ -806,7 +824,7 @@ func TestAppenderIndexFailedLogging(t *testing.T) {
 
 	core, observed := observer.New(zap.NewAtomicLevelAt(zapcore.DebugLevel))
 	indexer, err := docappender.New(client, docappender.Config{
-		FlushBytes: 500,
+		FlushBytes: 5000,
```
ericywl, line 827: Seems like this change fixed the issue: https://github.com/elastic/go-docappender/issues/211.

rubvs: I confirmed, this does seem to be the case.

PR #229, opened 2025-01-30
feat: add include_source_on_error param to bulk index requests

kruskall:
## ❓ Why is this being changed

At the moment ES might return the source a request in the error response. Set the parameter to avoid the fragile parsing logic and make bulk request handling more robust.
 
## 🧑‍💻 What is being changed

Set `IncludeSourceOnError` to `false`
 
 
## ✅ How to validate the change

unfortunately this can only be validate manually since it relies on an ES feature

### DIFF
--- a/appender.go
+++ b/appender.go
@@ -390,7 +390,7 @@ func (a *Appender) flush(ctx context.Context, bulkIndexer *BulkIndexer) error {
 		}
 
 		// Bulk indexing may fail with different status codes.
-		errFailed := ErrorFlushFailed{}
+		var errFailed ErrorFlushFailed
 		if errors.As(err, &errFailed) {
 			var legacy *int64
 			var status string


carsonip (commented):
Q: how does this interact with ES 8.x? Should we make this behavior in docappender either configurable, or maintain 2 docappender major versions for 8.x or 9.x ES separately?

bulk_indexer.go
```diff
@@ -433,19 +422,21 @@ func (b *BulkIndexer) newBulkIndexRequest(ctx context.Context) (*http.Request, e
...
 	v := req.URL.Query()
 	if b.config.Pipeline != "" {
 		v.Set("pipeline", b.config.Pipeline)
 	}
 	if b.config.RequireDataStream {
 		v.Set("require_data_stream", strconv.FormatBool(b.config.RequireDataStream))
 	}
-	v.Set("filter_path", "items.*._index,items.*.status,items.*.failure_store,items.*.error.type,items.*.error.reason")
+	v.Set("filter_path", strings.Join([]string{"items.*._index", "items.*.status", "items.*.failure_store", "items.*.error.type", "items.*.error.reason"}, ","))
+	v.Set("include_source_on_error", "false")
```
carsonip, line 439: Q: what happens when this is passed to a ES <8.18? If it will fail the request, doesn't it mean that elasticsearch exporter cannot bump docappender, as es exporter doesn't have strict dependency on ES version?

carsonip (changes requested):
It might present a problem with backward compatibility, as this code change introduces version dependency on ES, which may not be feasible for elasticsearchexporter.

A path forward may be to make this behavior configurable, to allow es exporter to opt out of this and use the manual sanaitization code path.

bulk_indexer.go
carsonip: it gives HTTP 400
```
POST /_bulk?include_source_on_error=true
{"create":{"_index":"logs-foo-bar"}}
{"@timestamp":"1", "foo":"bar"}
```

```json
{
  "error": {
    "root_cause": [
      {
        "type": "illegal_argument_exception",
        "reason": "request [/_bulk] contains unrecognized parameter: [include_source_on_error]"
      }
    ],
    "type": "illegal_argument_exception",
    "reason": "request [/_bulk] contains unrecognized parameter: [include_source_on_error]"
  },
  "status": 400
}
```

1pkg: I agree with Carson on this, there is a wider question about backward compatibility guarantees that we provide in this library. Since we version it independently from the stack version, do we need to always preserve backward compatibility?

simitt: >A path forward may be to make this behavior configurable, to allow es exporter to opt out of this and use the manual sanaitization code path.

You are raising an important point here @carsonip . While we could just bump to a major version after adding the change, it would make it much harder to use it and upgrade to the latest version downstream.
Making it configurable sounds like the right direction.

bulk_indexer.go
```diff
@@ -433,19 +422,21 @@ func (b *BulkIndexer) newBulkIndexRequest(ctx context.Context) (*http.Request, e
...
+	req.Header.Add("Content-Type", "application/json")
 	v := req.URL.Query()
 	if b.config.Pipeline != "" {
 		v.Set("pipeline", b.config.Pipeline)
 	}
 	if b.config.RequireDataStream {
 		v.Set("require_data_stream", strconv.FormatBool(b.config.RequireDataStream))
 	}
-	v.Set("filter_path", "items.*._index,items.*.status,items.*.failure_store,items.*.error.type,items.*.error.reason")
+	v.Set("filter_path", strings.Join([]string{"items.*._index", "items.*.status", "items.*.failure_store", "items.*.error.type", "items.*.error.reason"}, ","))
```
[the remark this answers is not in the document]
carsonip: does it incur runtime penalty (e.g. alloc)? What if we keep it as a global var?

bulk_indexer.go
carsonip: @kruskall has already agreed on making this configurable in another discussion. The remaining bit will be whether to 
1. keep the old sanitization logic in docappender; or
2. remove that logic and offload that to users (e.g. es exporter) requiring extra ES compatibility

bulk_indexer.go
```diff
@@ -501,27 +492,11 @@ func (b *BulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
...
-		if err := jsoniter.NewDecoder(res.Body).Decode(&er); err == nil {
-			er.Error.Reason = ""
-			er.Error.CausedBy.Reason = ""
-
-			b, _ := json.Marshal(&er)
-			s = string(b)
-		}
-
-		e := errorFlushFailed{resp: s, statusCode: res.StatusCode}
+		e := errorFlushFailed{resp: string(b), statusCode: res.StatusCode}
```
carsonip, line 499: if we're going with plan (2) where we remove sanitization and offload it to docappender users, the way `resp: string(b)` is passed and printed into a string in `errorFlushFailed.Error()` makes it very hard for docappender users to parse.

carsonip (commented):
1 pending question on `strings.Join`.

If we release this, imo we should bump the major version as we now require ES 8.18 by default.

bulk_indexer.go
simitt: @kruskall @carsonip could you clarify why you chose option 2? We were discussing backwards compatibilty here, so I would have gone with option 1 - keep the sanitization logic as long as the user doesn't specifically configure `include_source_on_error: true`. 

While the setting is now configurable, it still requires `8.18` as the parameter will also be sent as `false` if nothing is configured. How about using a `bool ptr` here and only setting the flag to `false` if configured. In combination with keeping the sanitization logic, this should be backwards compatible. Any other concerns with this that I am overlooking?

carsonip: > I would have gone with option 1 - keep the sanitization logic as long as the user doesn't specifically configure include_source_on_error: true.

I think you meant `include_source_on_error: false` as `include_source_on_error: false` removes the need to sanitize, and it is the default.

> While the setting is now configurable, it still requires 8.18 as the parameter will also be sent as false if nothing is configured. 

Correct, but we will skip the parameter if `include_source_on_error: true`, and it will not require 8.18.

I think the key point to discuss here is whether it is reasonable to put the maintenance burden of old sanitization on docappender instead of users (i.e. downstream libraries). I was under the impression that @kruskall prefers to keep docappender clean such that it has a (implicit) compatibility matrix closer to that of apm-server than es exproter.

carsonip: also, another point made by @kruskall is that we cannot have real `include_source_on_error: true` if the config is a `bool`, not a `*bool`.

There are 3 cases
1. pass `include_source_on_error: true` to ES, perform old sanitization
2. pass `include_source_on_error: true` to ES, no sanitization
3. pass `include_source_on_error: false` to ES

If docappender config is a bool, we need to decide between 1 and 2.

I'm open to making it a `*bool`, then it'll require no changes in es exporter, but it'll require opt-in from apm-server etc. This will be the most backward-compatible way, but it doesn't have the best safe default.

simitt: IMO changes should be backwards compatible with older, but supported ES versions. I'd like to propose 2 options:

Option 1)
* use a `*bool`; if nothing is configured for `include_source_on_error`, the current logic is kept, no parameter is sent, response is sanitized
* if `include_source_on_error: true|false`: the flag is set when sending a request to ES, no sanitization is applied. 

We document the changes and that it is recommended to set `include_source_on_error: false` as default for any ES versions `>=8.18`. 

Option 2)
* `include_source_on_error: false` is set as default
* expose another config option `configure_source_on_error` which defaults to `true` but users can set to `false` to keep backwards compatibility with older ES versions OR make the `include_source_on_error` an `enum` with a third option `unset`. 

From a usability and implementation perspective I prefer option 1, as it doesn't break default behaviour and is the cleanest. 

The APM Server needs to be compatible with newer and equal version numbers of Elasticsearch, hence we can configure `include_source_on_error: false`.

carsonip: > Option 1)

I don't love `*bool`. It works but could make test code and usage a bit strange to read.

> configure_source_on_error which defaults to true

Having a true default is not super ergonomic because we don't have a `DefaultOptions()` for caller to get default values. We rely on the zero value of the config struct (questionable, but that's the way it is). If we want a `true` default, we'll need to update all callers to call `DefaultOptions()` first.

> make the include_source_on_error an enum with a third option unset.

That's actually not a bad plan. It is not the simplest, but having `False` (default), `True`, `TrueButSanitize` will be quite descriptive. Obviously the naming could be better.

simitt: >configure_source_on_error which defaults to true

>Having a true default is not super ergonomic because we don't have a DefaultOptions() for caller to get default values. We rely on the zero value of the config struct (questionable, but that's the way it is). If we want a true default, we'll need to update all callers to call DefaultOptions() first.

We could work around it by negating the name of the second config option, e.g. `ignore_source_on_error_config`, the zero value is `false` and it would lead to the same result without requiring `DefaultOptions()`. It's hacky though to have two config options for the same thing, so I would also prefer an enum if we go with option (2); another suggestion for enum values `True | False | Sanitized`.

carsonip: enum sgtm. If we're exposing these under docappender package, users will be calling it `docappender.True` which is a bit strange. What about `IncludeSourceTrue`, `IncludeSourceFalse`, `IncludeSourceTrueSanitize`?

bulk_indexer.go
```diff
@@ -92,6 +91,15 @@ type BulkIndexerConfig struct {
 	//
 	// RequireDataStream is disabled by default.
 	RequireDataStream bool
+
+	// IncludeSourceOnError, if set to true, the response body of a Bulk Index request
+	// might contain the part of source document on error.
+	// Requires Elasticsearch 8.18+.
+	// WARNING: if set to true, user is responsible for sanitizing the error as it may contain
+	// sensitive data.
```
carsonip, line 99:
```suggestion
	// IncludeSourceOnError, if set to True, the response body of a Bulk Index request
	// might contain the part of source document on error.
	// Requires Elasticsearch 8.18+ if value is True or False.
	// WARNING: if set to True, user is responsible for sanitizing the error as it may contain
	// sensitive data.
```

config.go
```diff
@@ -113,6 +121,15 @@ type Config struct {
 	// RequireDataStream is disabled by default.
 	RequireDataStream bool
 
+	// IncludeSourceOnError, if set to true, the response body of a Bulk Index request
+	// might contain the part of source document on error.
+	// Requires Elasticsearch 8.18+.
+	// WARNING: if set to true, user is responsible for sanitizing the error as it may contain
+	// sensitive data.
```
carsonip, line 128:
```suggestion
	// IncludeSourceOnError, if set to True, the response body of a Bulk Index request
	// might contain the part of source document on error.
	// Requires Elasticsearch 8.18+ if value is True or False.
	// WARNING: if set to True, user is responsible for sanitizing the error as it may contain
	// sensitive data.
```

bulk_indexer.go
```diff
@@ -709,14 +729,22 @@ func indexnth(s []byte, nth int, sep rune) int {
 	})
 }
 
-type errorFlushFailed struct {
+type ErrorFlushFailed struct {
```
carsonip, line 732: this doesn't have to be exported now imo.

appender_test.go
```diff
@@ -773,86 +776,25 @@ func TestAppenderFlushRequestError(t *testing.T) {
...
-		test(t, http.StatusServiceUnavailable, "flush failed (503): {\"error\":{\"type\":\"x_content_parse_exception\",\"caused_by\":{\"type\":\"json_parse_exception\"}}}")
+		test(t, http.StatusServiceUnavailable, "flush failed (503): "+string(error))
 	})
 	t.Run("504", func(t *testing.T) {
-		test(t, http.StatusGatewayTimeout, "flush failed (504): {\"error\":{\"type\":\"x_content_parse_exception\",\"caused_by\":{\"type\":\"json_parse_exception\"}}}")
+		test(t, http.StatusGatewayTimeout, "flush failed (504): "+string(error))
 	})
 }
 
-func TestAppenderIndexFailedLogging(t *testing.T) {
```
carsonip, line 795: this test cannot be removed now. We need to check that reason is sanitized on unset

appender_test.go
```diff
@@ -58,6 +57,8 @@ func TestAppender(t *testing.T) {
 	var bytesTotal int64
 	var bytesUncompressed int64
 	client := docappendertest.NewMockElasticsearchClient(t, func(w http.ResponseWriter, r *http.Request) {
+		require.Len(t, r.URL.Query(), 1)
```
carsonip, line 60: it may sound stupid but I prefer we have a test that Unset, True, False are propagated correctly to the http request

appender_test.go
```diff
@@ -720,8 +722,9 @@ func TestAppenderFlushRequestError(t *testing.T) {
 		))
 
 		indexer, err := docappender.New(client, docappender.Config{
-			FlushInterval: time.Minute,
-			MeterProvider: sdkmetric.NewMeterProvider(sdkmetric.WithReader(rdr)),
+			FlushInterval:        time.Minute,
+			MeterProvider:        sdkmetric.NewMeterProvider(sdkmetric.WithReader(rdr)),
+			IncludeSourceOnError: docappender.True,
```
carsonip, line 727: As mentioned below, the Unset, False, True handling (as opposed to propagation) should be tested as well. False and True should pass through the error, Unset should sanitize reason.

bulk_indexer.go
```diff
@@ -709,14 +730,22 @@ func indexnth(s []byte, nth int, sep rune) int {
...
-func (e errorFlushFailed) Error() string {
+func (e *ErrorFlushFailed) StatusCode() int {
+	return e.statusCode
+}
+
+func (e *ErrorFlushFailed) ResponseBody() string {
+	return e.resp
+}
+
+func (e *ErrorFlushFailed) Error() string {
```
carsonip, line 749: nit: can we keep these as value receivers? The added benefit is we can skip the pointer gymnastics in returns and errors.As

bulk_indexer.go
carsonip: I commented because it isn't necessary in this PR, and it increases the public interface that we need to maintain in this lib. I'm fine with your current change.

appender_test.go
```diff
@@ -698,98 +700,94 @@ func TestAppenderFlushBytes(t *testing.T) {
...
-		var bytesTotal int64
-		var bytesUncompressedTotal int64
-		client := docappendertest.NewMockElasticsearchClient(t, func(w http.ResponseWriter, r *http.Request) {
-			bytesTotal += r.ContentLength
-			_, _, stat := docappendertest.DecodeBulkRequestWithStats(r)
-			bytesUncompressedTotal += stat.UncompressedBytes
-			w.WriteHeader(sc)
-			w.Write([]byte(`{"error": {"root_cause": [{"type": "x_content_parse_exception","reason": "reason"}],"type": "x_content_parse_exception","reason": "reason","caused_by": {"type": "json_parse_exception","reason": "reason"}},"status": 400}`))
-		})
+	for _, includeSource := range []docappender.Value{docappender.Unset, docappender.True} {
```
carsonip, line 710: non-blocker: I don't think we need to intentionally leave out docappender.False here. docappender.False will be the most used config and if we include it here, at least we know we're returning errors as-is for False and True.

bulk_indexer.go
```diff
@@ -537,6 +551,13 @@ func (b *BulkIndexer) Flush(ctx context.Context) (BulkIndexerResponseStat, error
 		return resp, fmt.Errorf("error decoding bulk response: %w", err)
 	}
 
+	if b.config.IncludeSourceOnError == Unset {
```
carsonip, line 554: nit: a comment here would be nice. In 1 year's time we'll forget why we sanitize on Unset but skipping sanitization on True / False. Or add the comment near the config.

carsonip (dismissed):
code lgtm, some nits

carsonip (approved):

bulk_indexer.go
1pkg: Will the introduction of `include_source_on_error` conflict with backward compatibility regardless? In https://github.com/elastic/go-docappender/pull/229#discussion_r1979842438 @carsonip confirmed that older ES version will reject a request with `include_source_on_error` option. So even if we go with 3 state configuration option for `include_source_on_error` either with `*bool` or `enum`, users will still have to know somehow to not use anything but the default value of this option when they target older ES.

carsonip: As stated in the config comment, both True and False requires 8.18+. It'll be fine in apm server, if apm server <8.18 doesn't bump docappender. In es exporter we'll stick with Unset for a while. What kind of backward compatibility are you referring to here?

https://github.com/elastic/go-docappender/pull/229/files#diff-c7cec697c2474f331487a79c052f6150eca5c9a2162a74094df972d831e8223aR98
```// Requires Elasticsearch 8.18+ if value is True or False.

1pkg: The worry here is that for an external library user it might be hard to discover this note, so in a hypothetical situation when someone updates this library and sets the flag with a good intention they will end up with a broken client.

I understand that setting safe default make it safe from backward compatibility perspective, but it kind of makes it more easy to misuse and claim that the library is not backward compatibility.

PR #231, opened 2025-02-11
Allow passing a custom item action

dmathieu:
When manipulating profiles data, we need to be able to use the `Update` action rather than `Create` all the time.
So this change allows passing custom actions rather than hardcoding create.

If no actions is provided, the default value will be `create`.

### DIFF
--- a/bulk_indexer.go
+++ b/bulk_indexer.go
@@ -58,7 +58,6 @@ const (
 	ActionUpdate = "update"
 )
 
-
 // BulkIndexerConfig holds configuration for BulkIndexer.
 type BulkIndexerConfig struct {
 	// Client holds the Elasticsearch client.
--- a/bulk_indexer.go
+++ b/bulk_indexer.go
@@ -304,7 +304,10 @@ func (b *BulkIndexer) Add(item BulkIndexerItem) error {
 }
 
 func (b *BulkIndexer) writeMeta(index, documentID, pipeline, action string, dynamicTemplates map[string]string) {
-	b.jsonw.RawString(`{"` + action + `":{`)
+	b.jsonw.RawString(`{"`)
+	b.jsonw.RawString(action)
+	b.jsonw.RawString(`":{`)
+
 	first := true
 	if documentID != "" {
 		b.jsonw.RawString(`"_id":`)
--- a/bulk_indexer.go
+++ b/bulk_indexer.go
@@ -280,13 +280,19 @@ type BulkIndexerItem struct {
 	Index            string
 	DocumentID       string
 	Pipeline         string
+	Action           string
 	Body             io.WriterTo
 	DynamicTemplates map[string]string
 }
 
 // Add encodes an item in the buffer.
 func (b *BulkIndexer) Add(item BulkIndexerItem) error {
-	b.writeMeta(item.Index, item.DocumentID, item.Pipeline, item.DynamicTemplates)
+	action := item.Action
+	if action == "" {
+		action = "create"
+	}
+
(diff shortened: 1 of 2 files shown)


bulk_indexer.go
```diff
@@ -297,8 +303,8 @@ func (b *BulkIndexer) Add(item BulkIndexerItem) error {
 	return nil
 }
 
-func (b *BulkIndexer) writeMeta(index, documentID, pipeline string, dynamicTemplates map[string]string) {
-	b.jsonw.RawString(`{"create":{`)
+func (b *BulkIndexer) writeMeta(index, documentID, pipeline, action string, dynamicTemplates map[string]string) {
+	b.jsonw.RawString(`{"` + action + `":{`)
```
carsonip, line 307:
```suggestion
	b.jsonw.RawString(`{"`)
	b.jsonw.RawString(action)
	b.jsonw.RawString(`":{`)
```
does this alloc? My instinct would be to split them into 3 calls

carsonip (commented):
nit on json encoding. Do you have everything you need in the bulk response? Asking since we designed bulk "indexer" to be specific to "indexing" new documents.

bulk_indexer.go
```diff
@@ -280,13 +291,23 @@ type BulkIndexerItem struct {
...
 func (b *BulkIndexer) Add(item BulkIndexerItem) error {
-	b.writeMeta(item.Index, item.DocumentID, item.Pipeline, item.DynamicTemplates)
+	action := item.Action
+	if action == "" {
+		action = ActionCreate
+	}
+
+	if !slices.Contains(allActions, action) {
+		return fmt.Errorf("%s is not a valid action", action)
+	}
```
carsonip, line 308:
```suggestion
	switch action {
	case ActionCreate, ActionDelete, ActionIndex, ActionUpdate:
	default:
		return fmt.Errorf("%s is not a valid action", action)
	}
```

carsonip (commented):
nit: perf overhead of slices.Contains

```
package slicescontains

import (
	"fmt"
	"slices"
	"testing"
)

const (
	// Actions are all the actions that can be used when indexing data.
	// `create` will be used by default.
	ActionCreate = "create"
	ActionDelete = "delete"
	ActionIndex  = "index"
	ActionUpdate = "update"
)

var allActions = []string{ActionCreate, ActionDelete, ActionIndex, ActionUpdate}

func BenchmarkSliceContains(b *testing.B) {
	var action = ActionCreate
	for i := 0; i < b.N; i++ {
		if !slices.Contains(allActions, action) {
			fmt.Errorf("%s is not a valid action", action)
		}
	}
}

func BenchmarkSwitch(b *testing.B) {
	var action = ActionCreate
	for i := 0; i < b.N; i++ {
		switch action {
		case ActionCreate, ActionDelete, ActionIndex, ActionUpdate:
		default:
			fmt.Errorf("%s is not a valid action", action)
		}
	}
}
```

```
goos: linux
goarch: amd64
pkg: playground/slicescontains
cpu: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz
BenchmarkSliceContains-16       295923946                4.132 ns/op           0 B/op          0 allocs/op
BenchmarkSwitch-16              1000000000               0.4741 ns/op          0 B/op          0 allocs/op
PASS
ok      playground/slicescontains       2.183s
```

carsonip (dismissed):

carsonip (approved):

rockdaboot (approved):
:+1:
Can we have new release?

PR #235, opened 2025-02-14
Add BulkIndexerItem.RequireDataStream

axw:
If set to true, require_data_stream=true
will be added to the bulk request item action.

### DIFF
--- a/bulk_indexer.go
+++ b/bulk_indexer.go
@@ -286,12 +286,13 @@ func (b *BulkIndexer) BytesUncompressedFlushed() int {
 }
 
 type BulkIndexerItem struct {
-	Index            string
-	DocumentID       string
-	Pipeline         string
-	Action           string
-	Body             io.WriterTo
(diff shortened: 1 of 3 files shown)


marclop:
#235 added support for this on a per-request level, is there an advantage to using this on a per-action level? I presume the answer is yes, but it seems like this will write a lot more bytes to the bulk indexer buffer.

axw:
@marclop sorry I should have added a link for context, this is motivated by https://github.com/open-telemetry/opentelemetry-collector-contrib/pull/37801 where we may want to set `require_data_stream` depending on the mapping mode (e.g. just for otel mode), or depending on the type of data (e.g. for certain profiling docs). We can set it at the request level, but that then forces data into different bulk requests.

We might not actually use this, but figured it wouldn't hurt to support it in go-docappender.

marclop (approved):
Looks good. Thanks for sharing the extra context.

PR #236, opened 2025-02-14
Add config PopulateFailedDocsSource to bulk indexer

carsonip:
Add config PopulateFailedDocsSource to bulk indexer to opt-in to getting back the source document that caused a failure. Should be used for testing and debugging only.


bulk_indexer.go
```diff
@@ -91,6 +91,12 @@ type BulkIndexerConfig struct {
 	//
 	// RequireDataStream is disabled by default.
 	RequireDataStream bool
+
+	// PopulateFailedDocsSource controls whether each BulkIndexerResponseItem.Source
+	// in BulkIndexerResponseStat.FailedDocs is populated with the source of the item,
+	// which includes the action line and the document line.
+	// For testing and debugging only. Use with caution.
+	PopulateFailedDocsSource bool
```
carsonip, line 99: [to reviewer] any suggestions for a better name? is `PopulateFailedDocsBody` clearer?

carsonip: Source might sound too much like `_source`, which is only the document content while in reality we return both action line and document line.

bulk_indexer.go
```diff
@@ -91,6 +91,13 @@ type BulkIndexerConfig struct {
 	//
 	// RequireDataStream is disabled by default.
 	RequireDataStream bool
+
+	// PopulateFailedDocsSource controls whether each BulkIndexerResponseItem.Source
+	// in BulkIndexerResponseStat.FailedDocs is populated with the source of the item,
+	// which includes the action line and the document line.
+	// For testing and debugging only. Use with caution as it may expose sensitive data.
+	// Any clients should relay this warning to users.
```
axw, line 99:
```suggestion
	//
	// WARNING: this is provided for testing and debugging only.
	// Use with caution as it may expose sensitive data; any clients
	// of go-docappender enabling this should relay this warning to
	// their users. Setting this will also add memory overhead.
```

Couple of suggestions:
 - make the warning a bit more prominent
 - mention that this will also impact memory usage (further discouraging use, and avoiding surprises)

bulk_indexer.go
axw: I initially assumed it was _source from the PR title, so I think it does sound a bit too much like that. Maybe `PopulateFailedDocsInput`?

bulk_indexer.go
```diff
@@ -132,6 +139,8 @@ type BulkIndexerResponseItem struct {
 		Type   string `json:"type"`
 		Reason string `json:"reason"`
 	} `json:"error,omitempty"`
+
+	Source string `json:"source"`
```
axw, line 143:
```suggestion
	Source string `json:"-"`
```

This doesn't come back from Elasticsearch, let's make that clear with `json:"-"`

bulk_indexer_test.go
```diff
@@ -295,3 +297,77 @@ func TestItemRequireDataStream(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, int64(2), stat.Indexed)
 }
+
+func TestPopulateFailedItemSource(t *testing.T) {
```
axw, line 301: Should we add a negative check, or is that covered elsewhere? i.e. check that Source is not populated unless we set the new config

axw (commented):
LGTM, just a few minor things

axw (dismissed):

bulk_indexer_test.go
```diff
@@ -295,3 +297,91 @@ func TestItemRequireDataStream(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, int64(2), stat.Indexed)
 }
+
+func TestPopulateFailedDocsInput(t *testing.T) {
+	test := func(enabled bool, compressionLevel int) {
```
1pkg, line 302: compressionLevel is unused

bulk_indexer_test.go
```diff
@@ -295,3 +297,91 @@ func TestItemRequireDataStream(t *testing.T) {
 	require.NoError(t, err)
 	require.Equal(t, int64(2), stat.Indexed)
 }
+
+func TestPopulateFailedDocsInput(t *testing.T) {
```
1pkg, line 301: Do we need a separate test that ensures that retry on failure branch with `PopulateFailedDocsInput` flag still works as expected for the `FailedDocs`?

carsonip:
Test is failing due to https://github.com/elastic/go-docappender/issues/241

bulk_indexer_test.go
carsonip: Done in https://github.com/elastic/go-docappender/pull/236/commits/451bb067c001968cb91f39b2cdbaed7b46c88f2a

carsonip (commented):
I have a temporary commit that changes the indent level such that the change is easier to review: https://github.com/elastic/go-docappender/pull/236/commits/f20b21e02e4a17a70e9b8671aa31e89bf9e57e96

axw (approved):

PR #238, opened 2025-02-27
feat: replace go-elasticsearch with elastic-transport-go

kruskall:
## ❓ Why is this being changed

we only use a bulk index request struct which can be safely replace since the body is passed manually.
This should significantly reduce the size of downstream clients that do not use go-elasticsearch.

Should unblock https://github.com/elastic/go-docappender/pull/229
 
## 🧑‍💻 What is being changed

- `esapi.Transport` -> `elastictransport.Interface`
- manually create the http request
 
## ✅ How to validate the change

go test ensures no change in behaviour. go-elasticsearch was already using elastic-transport under the hood

### DIFF
--- a/docappendertest/docappendertest.go
+++ b/docappendertest/docappendertest.go
@@ -60,13 +60,14 @@ type BulkIndexerResponse struct {
 
 // BulkIndexerResponseItem represents the Elasticsearch response item.
 type BulkIndexerResponseItem struct {
-	Index      string `json:"_index"`
-	DocumentID string `json:"_id"`
-	Version    int64  `json:"_version"`
-	Result     string `json:"result"`
-	Status     int    `json:"status"`
-	SeqNo      int64  `json:"_seq_no"`
-	PrimTerm   int64  `json:"_primary_term"`
+	Index        string `json:"_index"`
+	DocumentID   string `json:"_id"`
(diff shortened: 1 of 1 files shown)


rubvs (approved):

docappendertest/docappendertest.go
```diff
@@ -53,8 +52,41 @@ type BulkRequestItemMeta struct {
...
+		Cause  struct {
+			Type   string `json:"type"`
+			Reason string `json:"reason"`
+		} `json:"caused_by"`
+	} `json:"error,omitempty"`
+}
+
 // DecodeBulkRequest decodes a /_bulk request's body, returning the decoded documents and a response body.
-func DecodeBulkRequest(r *http.Request) ([][]byte, esutil.BulkIndexerResponse) {
+func DecodeBulkRequest(r *http.Request) ([][]byte, BulkIndexerResponse) {
```
carsonip, line 89: isn't this breaking the public interface? I'm fine with it but we'll expect a change in every project that uses docappender

carsonip (commented):
change looks fine. There is a breaking change in the docappendertest package. I wonder if we use Instrumentation of bulk request in the future, we need to revert this PR. But I don't see we're going to do that.

PR #243, opened 2025-03-25
Add BulkIndexerPool to improve resource sharing

### LINKED ISSUE #75: Introduce a `BulkIndexerPool` abstraction
Currently, the `Appender` creates `bulkIndexer`s at runtime when the Appender is created. Instead of doing that, we could pass in a `BulkIndexerPool`, which contains a bounded pool of `bulkIndexer`s and could be shared by multiple `Appender` instances to create a maximum memory boundary.

For fairness across different `go-docappender` instances, it would be good also to implement a per-go-docappender limit so one appended isn't potentially consuming all the bulk indexers.

marclop:
Adds a new `BulkIndexerPool` to the appender to improve resource sharing
between appenders, so they are reused and not created by each appender.
The pool can be shared between multiple instances of the `Appender` to
reduce overhead, resource sharing and improve performance.

The change is implemented in a backward compatible way.

The `BulkIndexerPool` accepts four parameters:

- `guaranteed` defines the guaranteed number of `BulkIndexer`s that can
be leased by a given ID. Even when the total limit is exceeded, more
`BulkIndexer`s can still be leased until `guaranteed` is hit.
- `max` defines the maximum number of `BulkIndexer`s that can be leased
by a given ID. It ensures that no more than this number can be active
at any time for a specific ID.
- `total` defines the total number of `BulkIndexer`s that can be leased
for the entire pool. It limits the overall resource usage of the pool.

Closes #75


appender.go
```diff
@@ -866,3 +827,15 @@ func timeFunc(f func()) time.Duration {
...
+
+// NOTE(marclop) until requiring 1.24.0 is possible, then we can switch to
+// rand.Text() from the `crypto/rand` package.
+func randStringBytes(length int) string {
+	buf := make([]byte, length)
+	for i := range buf {
+		buf[i] = alphanum[rand.Intn(len(alphanum))]
+	}
+	return string(buf)
+}
```
axw, line 841: Instead of generating a random ID, could we use the address of the Appender? i.e. `fmt.Sprintf("%p", a)`

config.go
```diff
@@ -76,6 +76,8 @@ type Config struct {
 	// If MaxRequests is less than or equal to zero, the default of 10 will be used.
 	MaxRequests int
 
+	BulkIndexerPool *BulkIndexerPool
```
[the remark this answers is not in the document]
axw: IIANM MaxRequests is now ignored if BulkIndexerPool is specified, is that right?

Would it make sense to make BulkIndexerPool optional? And if it's unset, create one dedicated to the Appender based on MaxRequests?

appender.go
```diff
@@ -153,33 +125,19 @@ func New(client elastictransport.Interface, cfg Config) (*Appender, error) {
...
 		config:    cfg,
-		available: available,
+		client:    client,
+		id:        randStringBytes(32),
 		closed:    make(chan struct{}),
 		bulkItems: make(chan BulkIndexerItem, cfg.DocumentBufferSize),
 		metrics:   ms,
 	}
-	indexer.addUpDownCount(int64(len(available)), &indexer.availableBulkRequests, ms.availableBulkRequests)
+	indexer.addUpDownCount(int64(cfg.MaxRequests), &indexer.availableBulkRequests, ms.availableBulkRequests)
```
axw, line 140: IIUC this means that every Appender will increment by MaxRequests, which I don't think is correct? Should we perhaps move this to the BulkIndexerPool?

config.go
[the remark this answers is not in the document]
axw: One more thought: to avoid the BulkIndexer.Flush API change, perhaps we could:
 - keep the Client in BulkIndexerConfig
 - ignore Client and Config.MaxRequests if and only if BulkIndexerPool is provided here -- otherwise construct one using the Client & Config.MaxRequests

WDYT?

bulk_indexer_pool.go
```diff
@@ -0,0 +1,209 @@
...
+	// that may be trying to acquire or release indexers.
+	// Even though this looks like it would prevent other Get() operations from
+	// proceeding, the lock is only held for a short time while we check the
+	// count and overall limits. The lock is released by:
+	// - p.cond.Wait() releases the lock while waiting for a signal.
+	// - p.mu.Unlock() releases the lock after the indexer is returned.
+	p.mu.Lock()
+	entry, exists := p.entries[id]
+	if !exists {
+		defer p.mu.Unlock()
```
axw, line 84: The distance between the locking here and unlocking in `p.get` worries me. Could we maybe refactor this a bit, to something like:

```go
func (p *BulkIndexerPool) Get(id string) *BulkIndexer {
	p.claim() // does the mutex/cond/atomics dance, but does not return a bulk indexer
	return p.getAvailable() // does everything get does now except the atomics & unlock
}
```

That does mean `Get` is slightly more expensive in the case where an entry doesn't yet exist, but that's only going to happen once, so I think the simplification would be worthwhile.

bulk_indexer_pool.go
```diff
@@ -0,0 +1,209 @@
...
+	p.cond = sync.NewCond(&p.mu)
+	return &p
+}
+
+// Get returns a BulkIndexer for the specified ID. If there aren't any existing
+// indexers for the ID, a new one is created.
+//
+// If the overall limit of indexers has been reached, it will wait until a slot
+// is available. This is a blocking operation.
+func (p *BulkIndexerPool) Get(id string) *BulkIndexer {
```
axw, line 73: Given that this is a blocking operation, should Get take a context and be cancellable?

axw (commented):
Looks great overall. I didn't look at the tests, and mostly focused on the interfaces.

appender.go
marclop: I think it'd be best to keep this here for the Appender and perhaps introduce a new metric (that doesn't depend on legacy metrics) in the `BulkIndexerPool`.

bulk_indexer_pool.go
marclop: I think we can just defer the `p.mu.Unlock`, then we don't need to refactor anything.

config.go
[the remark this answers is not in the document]
marclop: Addressed

bulk_indexer_pool.go
marclop: It is but at the moment is a bit awkward to pass any meaningful context in `appender.runActiveIndexer()`. The previous implementation didn't use any cancellations to receive from the channel and I'm not sure this one should? How do you see this one being used?

axw: Currently `runActiveIndexer` uses `active = <-a.available`, and the `a.available` channel is closed by the `Appender.Close` method. IIANM we could get into a state where other busy Appenders may block another one's `runActiveIndexer`. I think one way to address that would be:
 - add a BulkIndexerPool.Register method that creates an entry in `p.entries`
 - change BulkIndexer.Get to no longer create an entry if it does not exist; if it doesn't exist, it should return immediately, and similarly the loop should return early if it finds that the entry has been removed

I'm not 100% sure if it's necessary, just a bit suspicious whenever I see code that blocks but doesn't take a context.

marclop: good point, I had overlooked that case. Let me have a look and see what I can come up with.

bulk_indexer.go
```diff
@@ -286,6 +236,11 @@ func (b *BulkIndexer) Reset() {
 	b.bytesUncompFlushed = 0
 }
 
+// ResetClient resets the client used by the bulk indexer.
+func (b *BulkIndexer) ResetClient(client elastictransport.Interface) {
```
ericywl, line 240: I think `SetClient` is more appropriate here?

config.go
```diff
@@ -148,6 +161,53 @@ type Config struct {
 	MetricAttributes attribute.Set
 }
 
+// DefaultConfig returns a copy of cfg with any zero values set to their
+// default values.
+func DefaultConfig(cl elastictransport.Interface, cfg Config) Config {
```
ericywl, line 166: Minor nit: I would expect `DefaultConfig` to return a completely default `Config` instead of setting the empty values to default. Maybe something like this would be clearer:

```go
func (c *Config) initDefaults(cl elastictransport.Interface) {
    ...
}
```

bulk_indexer_pool.go
```diff
@@ -0,0 +1,207 @@
...
+	// proceeding, the lock is only held for a short time while we check the
+	// count and overall limits. The lock is released by:
+	// - p.cond.Wait() releases the lock while waiting for a signal.
+	// - p.mu.Unlock() releases the lock after the indexer is returned.
+	p.mu.Lock()
+	defer p.mu.Unlock()
+	entry, exists := p.entries[id]
+	if !exists {
+		entry, exists = p.entries[id]
+		if !exists {
```
ericywl, line 86: There seems to be 2 `!exists` check here? Since we are holding the mutex there shouldn't be a case where the `entries` get updated in-between.

bulk_indexer_pool.go
```diff
@@ -0,0 +1,207 @@
...
+// caller has already acquired the lock and checked the count and overall
+// limits. This function is called by Get() when the ID is already registered
+// and the overall limits have not been reached.
+func (p *BulkIndexerPool) get(entry idEntry) *BulkIndexer {
+	entry.count.Add(1)
+	p.leased.Add(1)
+	// First, try to return an existing non-empty indexer (if any). Try a few
+	// times since the mutex is unlocked and an indexer may have been returned
+	// in the meantime.
+	for i := 0; i < 3; i++ {
```
ericywl, line 132: Curious why the number `3` was chosen for the tries, instead of some other number or even some amount of time.

bulk_indexer_pool_test.go
```diff
@@ -0,0 +1,279 @@
...
+					}(indexer)
+				}
+				t.Log("end", id, pool.count(id))
+			}(id, p)
+		}
+		wg.Wait()
+		// Now we deregister all IDs and ensure that the pool is empty.
+		for id := range parameters {
+			c := pool.Deregister(id)
+			require.Equal(t, len(c), 0, id)
```
ericywl, line 100: Current usage swaps expected and actual, also for channel we can use:

```suggestion
			require.Len(t, c, 0, id)
```

bulk_indexer_pool_test.go
```diff
@@ -0,0 +1,279 @@
...
+			assert.Nil(t, <-c) // Assert nil (closed).
```
ericywl, line 101: I assume the channel will never have nils in it on normal circumstances, hence why we can check nil for closed? If so, ignore suggestion below:

```suggestion
            _, open := <-c
			assert.False(t, open) // Assert closed channel.
```

bulk_indexer_pool_test.go
```diff
@@ -0,0 +1,279 @@
...
+		pool.Put(key, bi)                           // Return to pool.
+
+		biC := pool.Deregister(key)
+		assert.NotNil(t, biC)
+		assert.Len(t, biC, 1)
+		var count int
+		for range biC {
+			count++
+		}
+		assert.Equal(t, 1, count)
```
ericywl, line 127: CMIIW but feels like this can be a separate test?

bulk_indexer_pool_test.go
```diff
@@ -0,0 +1,279 @@
...
+	indexers := make(chan *BulkIndexer, localMax)
+	for i := 0; i < localMax; i++ {
+		indexer, err := pool.Get(context.Background(), maxOutKey)
+		require.NoError(t, err)
+		require.NotNil(t, indexer)
+		indexers <- indexer
+	}
+
+	// Ensure that the local max is respected.
+	// This should block until the indexer is returned to the pool.
```
ericywl, line 175: Could be an outdated comment? I don't think `pool.count()` blocks.

bulk_indexer_pool_test.go
```diff
@@ -0,0 +1,279 @@
...
+	}()
+	select {
+	case <-started:
+		time.Sleep(20 * time.Millisecond) // Wait a bit.
+	case <-time.After(time.Second):
+		t.Fatal("Timed out waiting for goroutine to start")
+	}
+	// Ensure "test" key is maxed out.
+	assert.Equal(t, int64(localMax), pool.count(maxOutKey))
+	assert.Equal(t, pool.leased.Load(), int64(total))
```
ericywl, line 210: Since this one is a pre-condition to the test (instead of the actual thing that we are testing), I prefer using `require`.
```suggestion
	require.Equal(t, int64(localMax), pool.count(maxOutKey))
	require.Equal(t, int64(total), pool.leased.Load())
```

bulk_indexer_pool_test.go
```diff
@@ -0,0 +1,279 @@
...
+		for i := 0; i < localMax; i++ {
+			once.Do(func() { close(started) })
+			indexer, err := pool.Get(context.Background(), maxOutKey)
+			require.NoError(t, err)
+			require.NotNil(t, indexer)
+		}
+	}()
+	select {
+	case <-started:
+		time.Sleep(20 * time.Millisecond) // Wait a bit.
```
ericywl, line 204: Not sure what the wait is for?

bulk_indexer_pool_test.go
```diff
@@ -0,0 +1,279 @@
...
+	for i := 0; i < localMax; i++ {
+		pool.Put(maxOutKey, <-indexers)
+	}
+	select {
+	case <-done:
+	case <-time.After(time.Second):
+		t.Fatal("Timed out waiting for goroutine to finish")
+	}
+	assert.Equal(t, int64(localMax), pool.count(maxOutKey))
+	assert.Equal(t, pool.leased.Load(), int64(total))
```
ericywl, line 221: Minor nit:

```suggestion
	assert.Equal(t, int64(total), pool.leased.Load())
```

bulk_indexer_pool_test.go
```diff
@@ -0,0 +1,279 @@
...
+	cfg := BulkIndexerConfig{
+		Client: docappendertest.NewMockElasticsearchClient(t, func(http.ResponseWriter, *http.Request) {}),
+	}
+	t.Run("put empty id", func(t *testing.T) {
+		require.NoError(t, cfg.Validate())
+		pool := NewBulkIndexerPool(1, 2, 2, cfg)
+		assert.NotPanics(t, func() {
+			indexer, err := pool.Get(context.Background(), "test")
+			assert.Error(t, err)
+			pool.Put("", indexer)
```
ericywl, line 269: Should there be checks against empty id? Also the `indexer` here is `nil`.

ericywl (commented):
Hey @marclop, I'm practicing my code review skills on this PR since it looked interesting and decently complex 😄. Feel free to let me know if some of the comments are too nit-picky.

bulk_indexer_pool_test.go
marclop: Closing a channel returns the zero value of the channel type. In this case we have a `chan *BulkIndexer`, so when the channel is closed, it returns the zero value of `*BulkIndexer`, which is `nil`.

bulk_indexer_pool_test.go
marclop: The test ensures that if empty/nil are passed, we expect it to behave a certain way, i.e. not panic.

bulk_indexer_pool_test.go
marclop: Could you clarify what you mean? If it's just swapping `Equal` for `Len` I can do it, but there's little value.

bulk_indexer_pool_test.go
marclop: Added a comment.

bulk_indexer_pool_test.go
ericywl: The method parameter is `require.Equal(t, expected, actual)`, whereas current usage has them swapped.

marclop: I see, I'll update it.

bulk_indexer_pool_test.go
ericywl: Just realized that `Register` skips empty ID, ignore this comment 😄

bulk_indexer_pool_test.go
ericywl: Yes, I meant that if we were to do:
```
c <- nil
assert.Nil(t, <-c) // Also passes
```

But I think in this case we basically never pass nil into channel, so either way works.

axw (dismissed):
LGTM!

bulk_indexer_pool.go
```diff
@@ -0,0 +1,233 @@
...
+	p.mu.RLock()
+	defer p.mu.RUnlock()
+	entry, exists := p.entries[id]
+	if !exists {
+		return // unknown id, discard indexer
+	}
+	defer func() { // Always decrement the count and signal the condition.
+		entry.leased.Add(-1)
+		p.leased.Add(-1)
+		p.cond.Signal() // Signal waiting goroutines
```
1pkg, line 172: Should we `Broadcast` here as well (it might increase contention)? but because of the condition in https://github.com/elastic/go-docappender/pull/243/files#diff-b9dca0d168033864f1386c833c23ac23b1146e0e0d727af8e0bc918efaba3cbcR104 a random awaken gorutine won't be always guaranteed to acquire an indexer.

bulk_indexer_pool.go
```diff
@@ -0,0 +1,233 @@
...
+
+import (
+	"context"
+	"errors"
+	"sync"
+	"sync/atomic"
+)
+
+// ErrPoolUnknownID is returned when an ID is not found in the pool.
+var ErrPoolUnknownID = errors.New("bulk_indexer_pool: unknown id")
```
1pkg, line 28: Q: What is benefit of having this single error type explicitly vs implicitly lazily creating entry on demand when ID is unknown vs using something simple like boolean to signal if bulk indexer was returned or not? As far as I can see this is the only possible error in the pool + context cancelation error that serves a sanity purpose.

1pkg (approved):

1pkg (approved):
Great work!

bulk_indexer_pool.go
marclop: True, probably not a bad idea. I'll change it.

bulk_indexer_pool.go
marclop: fixed

axw (approved):

Issue #252, opened 2025-05-23
Improve wait in unit tests


rubvs:
The wait [loops](https://github.com/elastic/go-docappender/blob/fb133108473302daf0a3f2c5f39b30e762d1de2e/appender_test.go#L292) added in some unit tests can be improve using a channel.

PR #253, opened 2025-05-23
Add function hooks to config

rubvs:
Closes https://github.com/elastic/go-docappender/issues/252

Use function hooks to make unit tests more antifragile by leveraging Go concurrency primitives instead of waits, making the tests more expressive and deterministic.

- Replace [wait loop](https://github.com/elastic/go-docappender/blob/fb133108473302daf0a3f2c5f39b30e762d1de2e/appender_test.go#L119) in `TestAppender` with `OnConsume` hook.
- Replace [wait loop](https://github.com/elastic/go-docappender/blob/fb133108473302daf0a3f2c5f39b30e762d1de2e/appender_test.go#L293) in `TestAppenderRetry` with `OnFlush` hook.

### DIFF
--- a/appender_test.go
+++ b/appender_test.go
@@ -266,7 +266,7 @@ func TestAppenderRetry(t *testing.T) {
 		attribute.String("a", "b"), attribute.String("c", "d"),
 	)
 
-	consumed := make(chan struct{})
+	flushed := make(chan struct{})
 	indexer, err := docappender.New(client, docappender.Config{
 		FlushInterval:      time.Minute,
 		FlushBytes:         750, // this is enough to flush after 9 documents
@@ -274,9 +274,9 @@ func TestAppenderRetry(t *testing.T) {
 		MaxDocumentRetries: 1,   // to test the document retry logic
 		MeterProvider:      sdkmetric.NewMeterProvider(sdkmetric.WithReader(rdr)),
 		MetricAttributes:   indexerAttrs,
-		OnConsume: func() {
+		OnFlush: func() {
 			select {
-			case consumed <- struct{}{}:
+			case flushed <- struct{}{}:
 			default:
 			}
 		},
@@ -291,7 +291,7 @@ func TestAppenderRetry(t *testing.T) {
 	}
 
 	select {
-	case <-consumed:
+	case <-flushed:
 	case <-time.After(2 * time.Second):
 		t.Fatalf("timed out waiting for the active bulk indexer to send one bulk request")
 	}
--- a/appender_test.go
+++ b/appender_test.go
@@ -100,38 +100,32 @@ func TestAppender(t *testing.T) {
 		attribute.String("a", "b"), attribute.String("c", "d"),
 	)
 
+	consumed := make(chan struct{})
 	indexer, err := docappender.New(client, docappender.Config{
 		FlushInterval:    time.Minute,
 		MeterProvider:    sdkmetric.NewMeterProvider(sdkmetric.WithReader(rdr)),
 		MetricAttributes: indexerAttrs,
+		OnConsume: func() {
+			select {
+			case consumed <- struct{}{}:
+			default:
+			}
+		},
 	})
 
 	require.NoError(t, err)
 	defer indexer.Close(context.Background())
 
-	available := indexer.Stats().AvailableBulkRequests
 	const N = 10
 	for i := 0; i < N; i++ {
 		addMinimalDoc(t, indexer, "logs-foo-testing")
 	}
 
-	timeout := time.After(2 * time.Second)
-loop:
-	for {
-		select {
-		case <-time.After(10 * time.Millisecond):
-			// Because the internal channel is buffered to increase performance,
-			// the available indexer may not take documents right away, loop until
-			// the available bulk requests has been lowered.
-			if indexer.Stats().AvailableBulkRequests < available {
-				break loop
-			}
-		case <-timeout:
-			t.Fatalf("timed out waiting for the active bulk indexer to pull from the available queue")
-		}
+	select {
+	case <-consumed:
+	case <-time.After(2 * time.Second):
+		t.Fatalf("timed out waiting for the active bulk indexer to send one bulk request")
 	}
-	// Appender has not been flushed, there is one active bulk indexer.
-	assert.Equal(t, docappender.Stats{Added: N, Active: N, AvailableBulkRequests: 9, IndexersActive: 1}, indexer.Stats())
 
 	// Closing the indexer flushes enqueued documents.
 	err = indexer.Close(context.Background())
@@ -272,6 +266,7 @@ func TestAppenderRetry(t *testing.T) {
 		attribute.String("a", "b"), attribute.String("c", "d"),
 	)
 
+	consumed := make(chan struct{})
 	indexer, err := docappender.New(client, docappender.Config{
 		FlushInterval:      time.Minute,
 		FlushBytes:         750, // this is enough to flush after 9 documents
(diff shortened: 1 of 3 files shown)


marclop (commented):
I like that this improves test predictability, but I don't know if we have any use for these hooks (particularly with no arguments passed to them) outside of the unit tests.

rubvs:
@marclop, I was very hesitant to create this PR for exactly the usage reason you outlined. However,

- I find this the most clean implementation for the unit tests.
- Since this is a config addition, meaning it's optional, I think it is fine.
- This might inspire some use cases or refactors, since config hooks is a common pattern in worker pools.

I opened this PR on my hesitancy to get feedback, or maybe a better implementation idea for the unit tests.

axw (commented):
Similar comment to Marc, since this isn't used in the public API I'm very hesitant about this change.

Also, it's unclear to me how this is an improvement. What's wrong with the loop? FYI you can do the same thing that loop is doing with `assert.Eventually` which should help on the expressivity front.

rubvs:
@marclop @axw let me give some more context.

This change originated from another PR in which I'm removing the `Stats` datastructure, see https://github.com/elastic/go-docappender/pull/249. Here is my thinking.

1. Keep using the specific metric as a signal to wait.
2. Replace it with concurrency primitives to communicate state.
3. Compare (1) and (2). Make a decision.

## 1. Keep using the specific metric as a signal to wait.

Initially, I bruteforce swapped the `Stats` impl with that of the OTEL metrics while keeping the loop. I did actually replace the loop with `Eventually` as outlined in https://github.com/elastic/go-docappender/pull/253#pullrequestreview-2867055747. This was very cumbersome leeding me to have a deeper look.

## 2. Replace it with concurrency primitives to communicate state.

While initially hacking this change for improved understanding, I discovered a very flaky race in `TestAppenderRetry` that only sometimes happens on my Mac and not in CI:

1. The test leverages the `BulkRequests` metric to ensure waiting for a condition. The moment the flush function ends, the metric `BulkRequests` is `1`, breaking out of the loop, see [Code](https://github.com/elastic/go-docappender/blob/fb133108473302daf0a3f2c5f39b30e762d1de2e/appender_test.go#L300)
2. This metric is only incremented in one place. When the `flush` function ends, see [Code](https://github.com/elastic/go-docappender/blob/fb133108473302daf0a3f2c5f39b30e762d1de2e/appender.go#L283). The flush is called in a goroutine, see [Code](https://github.com/elastic/go-docappender/blob/fb133108473302daf0a3f2c5f39b30e762d1de2e/appender.go#L569).
4. If we communicate back to the test goroutine with a channel in this `defer`, we get a slight race (see **Reproduce Error** section below).
    - The test continues in one goroutine (the unit test), while the flush goroutine continues executing in another.
    - The race happens when https://github.com/elastic/go-docappender/blob/fb133108473302daf0a3f2c5f39b30e762d1de2e/appender_test.go#L359 in `TestAppenderRetry` asserts before the metric updates in https://github.com/elastic/go-docappender/blob/fb133108473302daf0a3f2c5f39b30e762d1de2e/appender.go#L573 This leads to the test assert failing due to the metric `AvailableBulkRequests` expected to be `1`, but is `0`.
    - This is why in my PR, communication back to the main routine - `TestAppenderRetry` in this case - happens after all metrics are updated and not in the `flush` defer along with the `BulkRequests` metric.

After a screenshare with @1pkg (thanks!), we were not exactly sure why this happens. Nevertheless, this indicates the fragility of using a metric to community concurrently. The current implementation is a derivative of communiting via shared memory, which goes against the Go concurrency philosophy IMHO.

## 3. Comparison

Leveraging a metric to ensure state should be a code smell:

- Observability primitives should not be "acted" on, only observed.
- The underlying reason for the loop is to wait for a condition between goroutines.
- What if the metric is delete or update for some practical reason?
- Or the position of the metric gets shifted within the codebase?

Currently, our unit tests are entangled with our observability layer. This is an abstraction leak IMO. We should only use metrics in our tests to assert (observe), and not for event consistency which introduces fragility.

## 4. Conclusion

I created this separate PR from the `Stats` changes, since share the same concern you all do. Thinking about it again, I think https://github.com/elastic/go-docappender/pull/253#issuecomment-2908424438 might be a good intermediate approach.

## Reproduce Error

- Apply the following patch to this PR's branch.
```diff
diff --git a/appender.go b/appender.go
index 5ed48f8..8f11d40 100644
--- a/appender.go
+++ b/appender.go
@@ -284,7 +284,12 @@ func (a *Appender) flush(ctx context.Context, bulkIndexer *BulkIndexer) error {
        if n == 0 {
                return nil
        }
-       defer a.addCount(1, &a.bulkRequests, a.metrics.bulkRequests)
+       defer func() {
+               a.addCount(1, &a.bulkRequests, a.metrics.bulkRequests)
+               if a.onFlush != nil {
+                       a.onFlush()
+               }
+       }()

        logger := a.config.Logger
        var span trace.Span
@@ -584,9 +589,6 @@ func (a *Appender) runActiveIndexer() {
                                        took.Seconds(),
                                        attrs,
                                )
-                               if a.onFlush != nil {
-                                       a.onFlush()
-                               }
                                return err
                        })
                        a.metrics.bufferDuration.Record(context.Background(),
```

- Run the unit test
```shell
> go test -v -race ./... -run TestAppenderRetry

=== RUN   TestAppenderRetry
    docappendertest.go:240:
                Error Trace:    /Users/rubenvanstaden/issue/go_docappender/252/go-docappender/docappendertest/docappendertest.go:240
                                                        /Users/rubenvanstaden/issue/go_docappender/252/go-docappender/appender_test.go:350
                                                        /Users/rubenvanstaden/issue/go_docappender/252/go-docappender/docappendertest/docappendertest.go:251
                                                        /Users/rubenvanstaden/issue/go_docappender/252/go-docappender/appender_test.go:333
                Error:          Not equal:
                                expected: 1
                                actual  : 0
                Test:           TestAppenderRetry
```

rubvs:
@axw I don't think the approach in https://github.com/elastic/go-docappender/pull/253#issuecomment-2910843152 will work. The issue arises with a race between the metrics being updated *after* the ES response and the asserts in the test routine.

I think adding a wait after the items have been added as in https://github.com/elastic/go-docappender/pull/249 is probably the most simplest approach. In practice there will always be a few second wait before we start collecting metrics. Therefore, we don't - and will probably never - see this race in practice.

My conclusion would still be to add a pure wait as in the `Stats` PR instead of the loop which adds confusion, but I'd like to get your final take on this before closing this PR.

rubvs:
In https://github.com/elastic/go-docappender/pull/253/commits/51e0e602334eed07479d473fd413c7144b7c3e3f, I moved the wait to after the ES response, which I think is a better approach.

appender_test.go
```diff
@@ -109,29 +115,11 @@ func TestAppender(t *testing.T) {
...
-			if indexer.Stats().AvailableBulkRequests < available {
-				break loop
-			}
-		case <-timeout:
-			t.Fatalf("timed out waiting for the active bulk indexer to pull from the available queue")
-		}
-	}
-	// Appender has not been flushed, there is one active bulk indexer.
-	assert.Equal(t, docappender.Stats{Added: N, Active: N, AvailableBulkRequests: 9, IndexersActive: 1}, indexer.Stats())
+	<-ch
```
axw, line 122: This blocks the test. There's no flush expected until the Close call below.

axw (commented):
> @axw I don't think the approach in #253 (comment) will work. The issue arises with a race between the metrics being updated after the ES response and the asserts in the test routine.

I don't follow. What I'm suggesting:
 1. wait until we know ES has responded
 2. verify metrics are updated

I _think_ what you're saying is there's a non-deterministic delay between ES responding and the metrics being updated. This is common asynchronous behaviour, and we can verify it by using the `assert.Eventually` pattern. As the function name suggests, we're verifying that, eventually, the metrics reflect the event that just occurred.

Can you elaborate on which part of this is not workable, or is undesirable? As @kruskall points out in his review, the change you made so far is going to be less robust than what we had.

PR #265, opened 2025-08-21
Introduce merge and split experimental APIs for bulk indexers

lahsivjar:
The merge and split APIs will allow Elasticsearch exporter to size the requests to ES using the new batch queue based on the bulk request sizes.

### DIFF
--- a/bulk_indexer.go
+++ b/bulk_indexer.go
@@ -18,9 +18,11 @@
 package docappender
 
 import (
+	"bufio"
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -256,6 +258,21 @@ func (b *BulkIndexer) resetBuf() {
 	}
 }
 
+// Size returns the size of the buffer used by bulk indexer as per
+// the specified sizer type.
+// EXPERIMENTAL: This is an experimental API
+func (b *BulkIndexer) Size(sizerType SizerType) int {
+	switch sizerType {
+	case ItemsCountSizer:
+		return b.Items()
+	case UncompressedBytesSizer:
+		return b.UncompressedLen()
+	case CompressedBytesSizer:
+		return b.Len()
+	}
+	return b.Items()
+}
+
 // Items returns the number of buffered items.
 func (b *BulkIndexer) Items() int {
 	return b.itemsAdded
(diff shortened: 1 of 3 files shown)


bulk_indexer.go
```diff
@@ -381,6 +399,128 @@ func (b *BulkIndexer) writeMeta(
...
+	switch b.config.CompressionLevel {
+	case gzip.NoCompression:
+		if _, err := other.buf.WriteTo(b.writer); err != nil {
+			return fmt.Errorf("failed to merge uncompressed bulk indexers: %w", err)
+		}
+	default:
+		// All compression levels
+		if other.gzipw != nil {
+			if err := other.gzipw.Close(); err != nil {
+				return fmt.Errorf("failed to merge compressed bulk indexers:%w", err)
```
rubvs, line 424:
```suggestion
				return fmt.Errorf("failed to merge compressed bulk indexers: %w", err)
```

bulk_indexer.go
```diff
@@ -381,6 +399,128 @@ func (b *BulkIndexer) writeMeta(
...
+//
+// EXPERIMENTAL: This is an experimental API and can be removed or
+// modified with breaking changes.
+func (b *BulkIndexer) Split(maxSize int, sizerType SizerType) ([]*BulkIndexer, error) {
+	size := b.Size(sizerType)
+	if size == 0 || size <= maxSize {
+		return []*BulkIndexer{b}, nil
+	}
+
+	var err error
```
rubvs, line 452: I don't think we need this `var err error` line.

bulk_indexer.go
```diff
@@ -381,6 +399,128 @@ func (b *BulkIndexer) writeMeta(
...
+		}
+
+		data, err := reader.ReadSlice('\n')
+		if err != nil {
+			return nil, fmt.Errorf("failed to split bulk requests, failed to read item: %w", err)
+		}
+		_, err = tmpBuffer.Write(data)
+		if err != nil {
+			return nil, fmt.Errorf("failed to split bulk requests, failed to write item: %w", err)
+		}
```
rubvs, line 501:
```suggestion
if _, err := tmpBuffer.Write(data); err != nil {
	return nil, fmt.Errorf("failed to split bulk requests, failed to write item: %w", err)
}
```

bulk_indexer.go
```diff
@@ -381,6 +399,128 @@ func (b *BulkIndexer) writeMeta(
...
+			}
+			return nil, fmt.Errorf("failed to split bulk requests, failed to read metadata: %w, %v", err, meta)
+		}
+		_, err = tmpBuffer.Write(meta)
+		if err != nil {
+			return nil, fmt.Errorf("failed to split bulk requests, failed to write metadata: %w", err)
+		}
+
+		data, err := reader.ReadSlice('\n')
+		if err != nil {
```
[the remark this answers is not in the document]
lahsivjar: Ah, I see your point - we should check for error not to be EOF here.

lahsivjar:
> @lahsivjar this is only going to be used by the otel elasticsearch exporter right? I wonder if we should expose some lower level functionality from go-docappender, and keep the split/merge logic in the exporter?

> Maybe one option for that would be to introduce a new "BulkIndexerBuffer" type that contains the Add method. It could have a Split method, or it could perhaps just have a method for iterating over items in the buffer and we can leave it to the caller to copy those to a new buffer.

Good idea! I will look into it.

lahsivjar:
@mauri870 It has taken a back seat due to different priorities but I have a working branch implementing @axw 's proposal which I will try to polish and send for a review early next week.

PR #273, opened 2025-09-11
perf: avoid copying body buffer when compression is enabled

kruskall:
slices.Grow is responsible for ~13% of inuse_space according to apm-server benchmarks.

The buffer copy is only used by the retry mechanism but it's performed regardless.

When compression is enabled, the default for apm‑server and most consumers of this library, the payload is streamed through a temporary slice so duplicating the original buffer isn’t needed.

Although the retry path works directly on the buffer when compression is disabled, focusing on the default (compressed) path lets us safely drop the `copybuf` slice and reuse the existing byte slice, thereby reducing memory usage.

### DIFF
--- a/bulk_indexer.go
+++ b/bulk_indexer.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"slices"
 	"strconv"
 	"unsafe"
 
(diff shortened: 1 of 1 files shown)


simitt:
@kruskall , https://github.com/elastic/go-docappender/pull/273#pullrequestreview-3217189664 is an important comment. I had not seen this PR before and share the concerns. Please elaborate before updating the version in any of our services and ensure this isn't getting released anywhere yet.

PR #274, opened 2025-09-12
perf: avoid copying body buffer when compression is enabled v2

kruskall:
Revert "perf: avoid copying body buffer when compression is enabled (https://github.com/elastic/go-docappender/pull/273)"
This reverts commit https://github.com/elastic/go-docappender/commit/33d964ca9a5a361aad8fc75fc0e76db681832465.

____

perf: avoid copying body buffer when compression is enabled

only copy the request body when a document fails, instead of doing so for every request.

Skip retry logic if there are no failed docs.

___

### DIFF
--- a/bulk_indexer.go
+++ b/bulk_indexer.go
@@ -24,7 +24,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"slices"
 	"strconv"
 	"unsafe"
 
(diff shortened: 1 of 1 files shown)


inge4pres (approved):

lahsivjar (commented):
LGTM!

